PR #2, opened 2024-01-31
Allow overriding KDIR and MODDIR in the makefile

ziguana:


ziguana:
You're right, thank you. Sorry, I'm not very good at make.

PR #3, opened 2024-02-01
Include the header for fixed width integer types

ziguana:
My distro is probably setting some pedantic compiler flags that fail the build without this header.


alewycky-tenstorrent:
Thanks for fixing this. I've merged your changes.

PR #4, opened 2024-02-07
Add a Nix module for NixOS users

jaykru:


alewycky-tenstorrent:
Sorry to take so long, I had to learn some NixOS.

I think tt-kmd.nix should be published on nixpkgs instead. Including it in the driver tree means that I have to push it after the release is made because it includes the directory hash. And then the version tagged with ttkmd-VER will always point to the previous version.

warthog9:
I should probably look this over when I get the rest of the distro package stuff sorted too, mostly commenting to remember I need to look at this later

PR #8, opened 2024-02-21
Draft: hwmon support for N300

joelsmithTT:
Adds hwmon support for N300 cards, where the second chip is not visible to Linux via PCIe but is connected to the first chip via ethernet. 

* Add simple TLB management mechanism
  * Introduces `tlb_t`: pattern is to allocate, use, and free a `tlb_t`
  * Solves the problem of multiple processes attempting to use the same TLB for different purposes
  * Functions for performing reads and writes to the region of BAR0 that corresponds to a `tlb_t`
  * Currently only manages the single TLB allocated to the kernel

* Small refactor of hwmon code
  * Introduce a layer of indirection to support both GS and WH with the same implementation in hwmon.c
  * Introduce notion of hwmon channels

* Introduce N300 hwmon support
  * Add code for probing the state of a local WH chip's ethernet cores
  * Add code for remote chip IO

Example `sensors` output:

```
grayskull-pci-a100
Adapter: PCI adapter
vcore:       740.00 mV (max =  +0.93 V)
asic_temp:    +25.3°C  (high = +75.0°C)
power:         3.00 W  (max = 170.00 W)
current:       6.00 A  (max = +300.00 A)

wormhole-pci-6100
Adapter: PCI adapter
vcore1:      720.00 mV (max =  +0.95 V)
vcore2:      720.00 mV (max =  +0.95 V)
asic_temp1:   +46.5°C  (high = +75.0°C)
asic_temp2:   +33.4°C  (high = +75.0°C)
power1:       12.00 W  (max =  85.00 W)
power2:       11.00 W  (max =  85.00 W)
current1:     15.00 A  (max = +160.00 A)
current2:     14.00 A  (max = +160.00 A)
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: GPL-2.0-only
 
 obj-m += tenstorrent.o
-tenstorrent-y := module.o chardev.o enumerate.o interrupt.o grayskull.o wormhole.o pcie.o hwmon.o
+tenstorrent-y := module.o chardev.o enumerate.o interrupt.o grayskull.o wormhole.o pcie.o hwmon.o eth.o tlb.o
 
 KDIR := /lib/modules/$(shell uname -r)/build
 KMAKE := $(MAKE) -C $(KDIR) M=$(CURDIR)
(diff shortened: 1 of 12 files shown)


joelsmithTT:
I have a different implementation of this for which I will open a different PR.

PR #9, opened 2024-02-27
hwmon: add ETH support for N300

joelsmithTT:


eth.c
```diff
@@ -0,0 +1,210 @@
...
+
+#define ETH_CMD_WR_REQ (0x1 << 0)
+#define ETH_CMD_WR_ACK (0x1 << 1)
+#define ETH_CMD_RD_REQ (0x1 << 2)
+#define ETH_CMD_RD_DATA (0x1 << 3)
+
+#define ETH_STATUS_UNKNOWN 0
+#define ETH_STATUS_NOT_CONNECTED 1
+
+static const u32 WH_ETH_NOC0_X[WH_ETH_CORE_COUNT] = { 9, 1, 8, 2, 7, 3, 6, 4, 9, 1, 8, 2, 7, 3, 6, 4 };
```
olofj, line 36: What do these constants mean? Might be useful with a one-line comment for context.

eth.c
```diff
@@ -0,0 +1,210 @@
...
+static const u32 WH_ETH_NOC0_Y[WH_ETH_CORE_COUNT] = { 0, 0, 0, 0, 0, 0, 0, 0, 6, 6, 6, 6, 6, 6, 6, 6 };
+
+struct eth_cmd_t {
```
olofj, line 39: Mind doing some one-liners to annotate here as well? I know it's probably just lifted from some existing header files, but we can aim to be a bit nicer to our users. Either oneliners here, or the standard kerneldoc format: https://docs.kernel.org/doc-guide/kernel-doc.html

eth.c
```diff
@@ -0,0 +1,210 @@
...
+		struct connected_eth_core *core_info = &wh_dev->connected_eth_cores[wh_dev->num_connected_cores];
+		u32 x = WH_ETH_NOC0_X[eth_channel];
+		u32 y = WH_ETH_NOC0_Y[eth_channel];
+		u32 fw_version = wh_noc_read32(tt, x, y, ETH_FW_VERSION_ADDR);
+		u32 port_status = wh_noc_read32(tt, x, y, ETH_PORT_STATUS_ADDR + (eth_channel * 4));
+		u32 remote_rack = wh_noc_read32(tt, x, y, ETH_REMOTE_RACK_ADDR);
+		u32 remote_shelf = wh_noc_read32(tt, x, y, ETH_REMOTE_SHELF_ADDR);
+		u32 rack_shelf = wh_noc_read32(tt, x, y, ETH_LOCAL_RACK_SHELF_ADDR);
+
+		if (fw_version < ETH_MIN_FW_VERSION)
```
olofj, line 93: Maybe a dev_info() here and below where you skip them?

eth.c
```diff
@@ -0,0 +1,210 @@
...
+	u32 y = WH_ETH_NOC0_Y[eth_channel];
+	u32 req_rd, req_wr, resp_rd, resp_wr; // queue pointers; 0 <= ptr < 8
+	u32 req_slot, resp_slot; // queue indices; 0 <= slot < 4
+	u32 req_offset, resp_offset;
+	u32 resp_flags_offset, resp_data_offset;
+	u32 fw_version;
+	u32 resp_flags = 0;
+
+	fw_version = wh_noc_read32(tt, x, y, ETH_FW_VERSION_ADDR);
+	if (fw_version < ETH_MIN_FW_VERSION) {
```
olofj, line 136: Do you need this on every read operation?

eth.c
```diff
@@ -0,0 +1,210 @@
...
+
+	// Encode the command.
+	memset(&cmd, 0, sizeof(struct eth_cmd_t));
+	cmd.sys_addr = encode_sys_addr(eth_addr->shelf_x, eth_addr->shelf_y, noc_x, noc_y, addr);
+	cmd.data = sizeof(u32);
+	cmd.rack = ((u16)eth_addr->rack_y) << 8 | eth_addr->rack_x;
+	cmd.flags = ETH_CMD_RD_REQ;
+
+	if (eth_queue_full(req_wr, req_rd)) {
+		pr_err("ETH queue %u full\n", eth_channel);
```
olofj, line 156: I think this can get very spammy? You might want to look into the ratelimit helpers.

eth.c
```diff
@@ -0,0 +1,210 @@
...
+			break;
+		}
+	}
+
+	// Busy wait until flags are set.
+	resp_slot = resp_rd & 3; // 0 <= resp_slot < 4
+	resp_offset = resp_slot * sizeof(struct eth_cmd_t);
+	resp_flags_offset = resp_offset + offsetof(struct eth_cmd_t, flags);
+	timeout = jiffies + msecs_to_jiffies(ETH_TIMEOUT_MS);
+	while ((resp_flags = wh_noc_read32(tt, x, y, ETH_RESP_QUEUE_ADDR + resp_flags_offset)) == 0) {
```
olofj, line 191: Are you sure you want to busywait here, instead of sleeping a little?

olofj (commented):
I didn't go through the whole diff yet, but meeting means I'd rather send this now than wait until I can do the rest.

alewycky-tenstorrent:
Sorry, I'm going to go back on something I said earlier. I said that we shouldn't use the fact that BAR4 happens to cover the kernel driver's TLB. But BAR0 can be reduced below 512MB in some systems, so BAR4 is the only assured way to access that TLB. Could you please use BAR4 instead? Conveniently, the last TLB is located at offset 0 in BAR 4.

alewycky-tenstorrent (changes requested):

wormhole.c
```diff
@@ -94,31 +119,98 @@ static bool wormhole_init(struct tenstorrent_device *tt_dev) {
...
+	{ hwmon_power, hwmon_power_max,   0x80, 16, GENMASK(15, 0), 1000000, 1 },
 	{ .reg_offset = TT_HWMON_ATTR_END },
 };
 
 static const struct tt_hwmon_label wh_hwmon_labels[] = {
-	{ hwmon_temp,  hwmon_temp_label,  "asic1_temp" },
-	{ hwmon_in,    hwmon_in_label,    "vcore1"     },
-	{ hwmon_curr,  hwmon_curr_label,  "current1"   },
-	{ hwmon_power, hwmon_power_label, "power1"     },
+	{ hwmon_temp,  hwmon_temp_label,  "asic_temp1", 0 },
```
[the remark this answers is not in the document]
joelsmithTT: "asic1" on N150 ends up being easier than "asic" (avoids having separate tables and sidesteps the problem of a mixed N150/N300 configuration where fixing up a single table would be incorrect).

So for the sake of any tooling that consumes the labels, I'll make GS have the `1` in the label names too.

wormhole.c
```diff
@@ -36,11 +38,33 @@
...
+#define TLB_SIZE_2M (2 * (1024 * 1024))
+#define TLB_SIZE_16M (16 * (1024 * 1024))
+#define TLB_COUNT_1M 156
+#define TLB_COUNT_2M 10
+#define TLB_COUNT_16M 20
+#define TLB_COUNT (156 + 10 + 20)
+#define TLB_1M_BASE 0
+#define TLB_2M_BASE (TLB_1M_BASE + TLB_SIZE_1M * TLB_COUNT_1M)
+#define TLB_16M_BASE (TLB_2M_BASE + TLB_SIZE_2M * TLB_COUNT_2M)
+#define TLB_OFFSET(index)                                                                             \
```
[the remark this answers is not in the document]
joelsmithTT: Not anymore

wormhole.c
```diff
@@ -94,31 +110,98 @@ static bool wormhole_init(struct tenstorrent_device *tt_dev) {
...
+	offset += ARC_CSM_NOC;
+
+	// Use the first connected core to determine the remote chip's address.
+	remote_addr = &wh_dev->connected_eth_cores[0].remote;
+
+	// Use eth core 0 to do the actual read.
+	eth_channel = 0;
+	lock_index = TENSTORRENT_LOCK_ETH(eth_channel);
+
+	// FIXME: UMD must use the KMD locking API to avoid conflicts here.
```
joelsmithTT, line 146: I need to address this in UMD before this gets merged.

joelsmithTT: There was a simple solution to this (use a core that UMD does not touch) that won't be viable due to an upcoming change for Metal: ETH cores that are not connected to a peer can not be relied on to have traditional firmware.  Although the first 8KB of L1 will be the same so logic to discover the topology will still work, software can't use any arbitrary core for remote transfers.

eth.c
```diff
@@ -0,0 +1,233 @@
...
+ * @noc_y: Y NOC coordinate of the core on the remote chip.
+ * @offset: Local offset within the remote chip's core.
+ *
+ * This function is Wormhole-specific and does not encode rack X/Y coordinates.
+ *
+ * Return: Encoded remote address, suitable for use in an eth_cmd_t.
+ */
+static u64 encode_sys_addr(u32 chip_x, u32 chip_y, u32 noc_x, u32 noc_y, u64 offset)
+{
+	u64 result = chip_y; // shelf
```
alewycky-tenstorrent, line 80: Could you try a bitfield+union and see if it comes out cleaner?

eth.c
```diff
@@ -0,0 +1,234 @@
...
+
+		wh_dev->num_connected_cores++;
+	}
+}
+
+bool wormhole_remote_read32(struct wormhole_device *wh_dev, u32 eth_channel, struct eth_addr_t *eth_addr, u32 noc_x,
+			    u32 noc_y, u64 addr, u32 *value)
+{
+	struct tenstorrent_device *tt = &wh_dev->tt;
+	struct sys_addr_t sys_addr;
```
alewycky-tenstorrent, line 143: Please initialize with = { 0 }; so the reserved bits are cleared.

wormhole.c
alewycky-tenstorrent: Is there a solution for safe queue submission? Other than this issue, I think this is ready to go.

olofj: I wonder if we can track whether we've seen that API being used by the processes that have the device open. If they've never used it, maybe we should just stay away and not access the remote. I.e. return -EBUSY where appropriate -- unless that triggers a retry/wait loop up the call chain that wouldn't complete.

eth.c
```diff
@@ -0,0 +1,232 @@
+// SPDX-FileCopyrightText: © 2024 Tenstorrent Inc.
+// SPDX-License-Identifier: GPL-2.0-only
+
+#define pr_fmt(fmt) KBUILD_MODNAME ": " fmt
```
olofj, line 4: nit: with dev_*() print functions, modname probably isn't useful since the device name will be in the prefix

wormhole.c
joelsmithTT: @alewycky-tenstorrent - the only way is a locking scheme (or don't share the core).
@olofj - interesting idea - I'll try that

joelsmithTT: @olofj, a first pass at implementing your idea is at 9c64bf82be55126aafa61d1bc7706d3f34964f96.

Here is what `sensors` looks like when the driver is returning -EBUSY in this case:
```
wormhole-pci-6100
Adapter: PCI adapter
asic1_vcore:   720.00 mV (max =  +0.95 V)
ERROR: Can't get value of subfeature in1_max: Can't read
asic2_vcore:        N/A  (max =  +0.00 V)
asic1_temp:     +47.2°C  (high = +75.0°C)
ERROR: Can't get value of subfeature temp2_max: Can't read
asic2_temp:         N/A  (high =  +0.0°C)
asic1_power:    12.00 W  (max =  85.00 W)
ERROR: Can't get value of subfeature power2_max: Can't read
asic2_power:        N/A  (max =   0.00 W)
asic1_current:  17.00 A  (max = +160.00 A)
ERROR: Can't get value of subfeature curr2_max: Can't read
asic2_current:      N/A  (max =  +0.00 A)
```

joelsmithTT: Improved implementation at 1e374252f14eaaa4f59e99c452b14078dd75ef28.
As an aside, hwmon indexing starts at 1 for everything but voltage (spelled `in`), which starts at 0 -- hence `in1_max` vs `temp2_max`, `power2_max`, etc. above.

alewycky-tenstorrent (approved):

mikey:
If I manually merge this with master it runs CI and the CI passes. 👍 
https://github.com/mikey/tt-kmd/actions/runs/9021603861

PR #12, opened 2024-03-21
Explicit type cast to work around compiler error during build in Arch Linux's DKMS hook

marty1885:
Hi,

I know this is an unsupported setup. I'm working on a set of AUR packages to install Tenstorrent software onto Arch Linux. This patch fixes test building during Arch processing the DKMS hook during a kernel or driver upgrade. Otherwise the compiler complains about `TENSTORRENT_MAX_DMA_BUFS` being casted to 0 due to the 3rd parameter in `AllocateDmaBuf` is of type `uint8_t` but the value of TENSTORRENT_MAX_DMA_BUFS is 256.


alewycky-tenstorrent:
Which compiler are you using and is it with -Werror?

Does changing "if" to "if constexpr" also fix the error for you?
```
    if constexpr (TENSTORRENT_MAX_DMA_BUFS <= std::numeric_limits<decltype(tenstorrent_allocate_dma_buf_in::buf_index)>::max()) {
        auto buf_max = AllocateDmaBuf(dev_fd, page_size(), TENSTORRENT_MAX_DMA_BUFS);
```

PR #18, opened 2024-04-26
Add initial github CI and .gitignore

mikey:
This does make and dkms based builds on a bunch of ubuntu versions

Also produces build artifacts for people to use. 

I've not tested the build aritfacts.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,37 @@
+name: 'test'
+
+on:
+  push:
+  pull_request:
+  workflow_dispatch:
+
+jobs:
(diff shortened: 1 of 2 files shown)


mikey:
This currently passes CI as can be seen here: 
https://github.com/mikey/tt-kmd/actions/runs/8841279445
We'll need to enable CI on this repo to see the same here.

mikey:
Updated to v4 checkout and aritfacts to avoid deprecation warnings from github actions. 

Clean run now.
https://github.com/mikey/tt-kmd/actions/runs/8842543917

PR #20, opened 2024-05-03
Establish pattern for exposing attributes in sysfs

joelsmithTT:
Expose card identification information and clock speeds.

Such information does not fit into the hwmon framework.

### DIFF
--- a/enumerate.c
+++ b/enumerate.c
@@ -81,6 +81,7 @@ static int tenstorrent_pci_probe(struct pci_dev *dev, const struct pci_device_id
 	pci_enable_pcie_error_reporting(dev);
 
 	pci_set_drvdata(dev, tt_dev);
+	dev_set_drvdata(&tt_dev->dev, tt_dev);
 
 	tt_dev->interrupt_enabled = tenstorrent_enable_interrupts(tt_dev);
 
@@ -96,12 +97,25 @@ static int tenstorrent_pci_probe(struct pci_dev *dev, const struct pci_device_id
 		register_reboot_notifier(&tt_dev->reboot_notifier);
 	}
 
+	if (tt_dev->attributes) {
+		struct tt_attribute_data *data = tt_dev->attributes;
+		for (; data->attr.attr.name; data++)
+			device_create_file(&tt_dev->dev, &data->attr);
+	}
+
 	return 0;
 }
 
 static void tenstorrent_pci_remove(struct pci_dev *dev)
 {
 	struct tenstorrent_device *tt_dev = pci_get_drvdata(dev);
+
+	if (tt_dev->attributes) {
+		struct tt_attribute_data *data = tt_dev->attributes;
+		for (; data->attr.attr.name; data++)
+			device_remove_file(&tt_dev->dev, &data->attr);
+	}
+
 
 	// These remove child sysfs entries which must happen before remove returns.
 	tenstorrent_unregister_device(tt_dev);
--- a/grayskull.c
+++ b/grayskull.c
@@ -135,6 +135,18 @@ struct TLB_16M_REG {
 
 #define TLB_16M_SIZE_SHIFT 24
 
+
+static struct tt_attribute_data gs_attributes[] = {
+	{ __ATTR(tt_aiclk,  S_IRUGO, tt_show_attribute, NULL), 0x50, 0xFFFF },
+	{ __ATTR(tt_axiclk, S_IRUGO, tt_show_attribute, NULL), 0x54, 0xFFFF },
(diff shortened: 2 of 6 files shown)


hwmon.c
```diff
@@ -69,3 +70,65 @@ const struct hwmon_ops tt_hwmon_ops = {
...
+	u64 card_type = (board_id_hi >> 4) & 0xFFFF;
+	if (card_type == 0x3)
+		return scnprintf(buf, PAGE_SIZE, "e150\n");
+	if (card_type == 0x7)
+		return scnprintf(buf, PAGE_SIZE, "e75\n");
+	if (card_type == 0x14)
+		return scnprintf(buf, PAGE_SIZE, "n300\n");
+	if (card_type == 0x18)
+		return scnprintf(buf, PAGE_SIZE, "n150\n");
+	return scnprintf(buf, PAGE_SIZE, "unknown\n");
```
alewycky-tenstorrent, line 98: Switch statement to pick string.

grayskull.c
```diff
@@ -135,6 +135,29 @@ struct TLB_16M_REG {
 
 #define TLB_16M_SIZE_SHIFT 24
 
+
+static struct tt_attribute_data gs_attributes[] = {
```
alewycky-tenstorrent, line 139: Can you just put a pointer into tenstorrent_device_class and move init/cleanup to device-indepdent code? Terminate the array with __ATTR_NULL.

hwmon.c
```diff
@@ -69,3 +70,65 @@ const struct hwmon_ops tt_hwmon_ops = {
...
+	struct tenstorrent_device *tt_dev = dev_get_drvdata(dev);
+	struct tt_attribute_data *data = container_of(attr, struct tt_attribute_data, attr);
+	struct tt_hwmon_context *ctx = &tt_dev->hwmon_context;
+
+	u32 board_id_hi = ioread32(ctx->telemetry_base + data->reg_offset);
+	u32 board_id_lo = ioread32(ctx->telemetry_base + data->reg_offset + 4);
+	return scnprintf(buf, PAGE_SIZE, "%08X%08X\n", board_id_hi, board_id_lo);
+}
+
+ssize_t tt_show_arc_fw(struct device *dev, struct device_attribute *attr, char *buf) {
```
alewycky-tenstorrent, line 111: Needs rename because it's used for other-than-arc fw.

enumerate.c
alewycky-tenstorrent, line 1: Ideally break into separate commit so the commit message can explain the change.

joelsmithTT: If you were referring to the reordering of some function calls in enumerate.c -- that turned out not to be needed.

hwmon.c
alewycky-tenstorrent: I was thinking more like
```
switch (card_type) {
    case 0x03: card_name = "e150"; break;
...
    default: card_name = "unknown"; break;
}

return scnprintf(buf, PAGE_SIZE, "%s\n", card_name);
```

enumerate.c
```diff
@@ -96,13 +97,30 @@ static int tenstorrent_pci_probe(struct pci_dev *dev, const struct pci_device_id
 		register_reboot_notifier(&tt_dev->reboot_notifier);
 	}
 
+	if (tt_dev->attributes) {
+		struct tt_attribute_data *data = tt_dev->attributes;
+		while (data->attr.attr.name) {
+			device_create_file(&tt_dev->dev, &data->attr);
+			data++;
+		}
```
alewycky-tenstorrent, line 105: I would prefer if the iteration is in a for loop like `for (; data->attr.attr.name; data++) {`. Ditto in tenstorrent_pci_remove.

hwmon.c
joelsmithTT: Fixed

hwmon.c
```diff
@@ -69,3 +70,66 @@ const struct hwmon_ops tt_hwmon_ops = {
...
+	return sprintf(buf, "%u\n", value);
+}
+
+ssize_t tt_show_card_type(struct device *dev, struct device_attribute *attr, char *buf) {
+	struct tenstorrent_device *tt_dev = dev_get_drvdata(dev);
+	struct tt_attribute_data *data = container_of(attr, struct tt_attribute_data, attr);
+	struct tt_hwmon_context *ctx = &tt_dev->hwmon_context;
+
+	u32 board_id_hi = ioread32(ctx->telemetry_base + data->reg_offset);
+	u64 card_type = (board_id_hi >> 4) & 0xFFFF;
```
alewycky-tenstorrent, line 89: Why is card_type u64? Seems it can't be more than 16 bits.

joelsmithTT: Fixed

alewycky-tenstorrent (approved):

PR #31, opened 2024-09-05
Fix ioctl input & output sizes

### LINKED ISSUE #30: ioctl bugs
1. The [`in` variable in `ioctl_get_device_info`](https://github.com/tenstorrent/tt-kmd/blob/3ae2a24997bee3ac91ce5144dd1fa4708831ca8a/chardev.c#L113) is declared as type `struct tenstorrent_get_device_info_out`, but presumably it wants to be `struct tenstorrent_get_device_info_in`.
2. The [`copy_to_user` call in `ioctl_get_device_info`](https://github.com/tenstorrent/tt-kmd/blob/3ae2a24997bee3ac91ce5144dd1fa4708831ca8a/chardev.c#L133-L135) currently uses `sizeof(out)` whereas it looks very much like it wants to use the immediately previously computed `bytes_to_copy = min(in.output_size_bytes, (u32)sizeof(out));`. 

Looks like `ioctl_get_driver_info` has the same two problems, meanwhile `ioctl_reset_device` gets the 1<sup>st</sup> correct but still has the 2<sup>nd</sup> wrong.

alewycky-tenstorrent:
Fix #30 with tests that verify the ioctl input & output data transfers are implemented correctly.

### DIFF
--- a/test/ioctl_zeroing.cpp
+++ b/test/ioctl_zeroing.cpp
@@ -1,0 +1,107 @@
+// Some ioctls have an output_size_bytes input value. When the actual output
+// data is smaller than output_size_bytes, the remainder must be zero-filled.
+
+#include <algorithm>
+#include <memory>
+#include <string>
+#include <vector>
+
+#include <cerrno>
+#include <cstddef>
+#include <cstdint>
+#include <cstdlib>
+
+#include <sys/ioctl.h>
+
+#include "aligned_allocator.h"
+#include "devfd.h"
+#include "enumeration.h"
+#include "ioctl.h"
+#include "test_failure.h"
+#include "util.h"
+
+namespace
+{
+
+#define CHECK_IOCTL_ZEROING(fd, ioctl_name, ioctl_data) CheckIoctlZeroing(fd, ioctl_name, #ioctl_name, ioctl_data)
+
+template <class IoctlData>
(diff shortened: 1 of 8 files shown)


test/ioctl_zeroing.cpp
```diff
@@ -0,0 +1,107 @@
...
+namespace
+{
+
+#define CHECK_IOCTL_ZEROING(fd, ioctl_name, ioctl_data) CheckIoctlZeroing(fd, ioctl_name, #ioctl_name, ioctl_data)
+
+template <class IoctlData>
+void CheckIoctlZeroing(int fd, unsigned long ioctl_code, const char *ioctl_name, const IoctlData &ioctl_data)
+{
+    std::vector<unsigned char, AlignedAllocator<unsigned char, alignof(IoctlData)>> buf;
+    buf.resize(sizeof(IoctlData::in) + page_size(), 0xFF);
```
joelsmithTT, line 32: Why `sizeof(IoctlData::in)` and not `sizeof(IoctlData)`?

joelsmithTT (approved):
LGTM.  I'm surprised I never noticed the double _out thing.

test/ioctl_zeroing.cpp
alewycky-tenstorrent: Because in.output_size_bytes is how many bytes are available for output starting at &arg->out, i.e sizeof(IoctlData::in) bytes into the structure.

But there could be padding between in & out, so actually it must be offsetof(IoctlData, out).

Issue #34, opened 2025-01-09
Null pointer dereference on process exit after device removed


alewycky-tenstorrent:
1. System setup: Wormhole, IOMMU enabled, Ubuntu 22.04 or 24.04, Kernel 5.15 or 6.8, driver version 1.30 or 1.31
2. Start a program that opens a device and allocates a DMA buffer. e.g. run drt1 (source to follow) with the target device, wait for "Ready"
3. In another terminal, remove the device by its PCI BDF `echo 1 | sudo tee /sys/bus/pci/devices/dddd:BB:DD:F/remove`
4. Back in the drt1 terminal press enter. It will exit and bash displays "Killed".
5. Check the system log (e.g. sudo journalctl -ke -p7) for errors.

No Oops occurs if the process only has an open fd but no DMA buffer. No Oops occurs if `iommu=pt` is on the kernel command line.

```
Jan 09 20:04:27 yyz-lab-77 kernel: BUG: kernel NULL pointer dereference, address: 00000000000000a8
Jan 09 20:04:27 yyz-lab-77 kernel: #PF: supervisor read access in kernel mode
Jan 09 20:04:27 yyz-lab-77 kernel: #PF: error_code(0x0000) - not-present page
Jan 09 20:04:27 yyz-lab-77 kernel: PGD 0 P4D 0
Jan 09 20:04:27 yyz-lab-77 kernel: Oops: 0000 [#1] PREEMPT SMP NOPTI
Jan 09 20:04:27 yyz-lab-77 kernel: CPU: 16 PID: 60080 Comm: drt1 Tainted: P           OE      6.8.0-51-generic #52-Ubuntu
Jan 09 20:04:27 yyz-lab-77 kernel: Hardware name: ASUS System Product Name/Pro WS WRX80E-SAGE SE WIFI, BIOS 1501 05/06/2024
Jan 09 20:04:27 yyz-lab-77 kernel: RIP: 0010:iommu_get_dma_domain+0xd/0x20
Jan 09 20:04:27 yyz-lab-77 kernel: Code: a8 6c 00 66 2e 0f 1f 84 00 00 00 00 00 90 90 90 90 90 90 90 90 90 90 90 90 90 90 90 90 0f 1f 44 00 00 55 48 8b 87 e0 02 00 00 <48> 8b 80 a8 00 00 00 48>
[...]

PR #36, opened 2025-01-23
Attempt to fix NULL pointer dereference

joelsmithTT:
An attempt to fix https://github.com/tenstorrent/tt-kmd/issues/34.

#34 says, "This presumably also affects pinned system memory" -- yes, I've run into this a couple of times.

I am allocating and mapping/pinning a large buffer with KMD and making it accessible to an L2CPU.  If I forget to quit the userspace process that does this before removing the card from the PCI bus, I get a similar NULL pointer dereference when I do kill the process.

I am removing the card from the bus so I can power it off when I'm not using it, while leaving the host system online.  (The particular card I am working with is in an eGPU dock with its own PSU).  `echo 1 | sudo tee /sys/bus/pci/devices/dddd:BB:DD:F/remove` and `echo 1 | sudo tee /sys/bus/pci/rescan` is a convenient mechanism for this.

### DIFF
--- a/chardev_private.h
+++ b/chardev_private.h
@@ -33,6 +33,8 @@ struct chardev_private {
 	struct list_head peer_mappings; // struct peer_resource_mapping.list
 
 	DECLARE_BITMAP(resource_lock, TENSTORRENT_RESOURCE_LOCK_COUNT);
+
+	struct list_head node;	// for struct tenstorrent_device::open_fds
 };
 
 struct chardev_private *get_tenstorrent_priv(struct file *f);
--- a/chardev_private.h
+++ b/chardev_private.h
@@ -34,7 +34,7 @@ struct chardev_private {
 
 	DECLARE_BITMAP(resource_lock, TENSTORRENT_RESOURCE_LOCK_COUNT);
 
-	struct list_head node;	// for struct tenstorrent_device::open_fds
+	struct list_head open_fd;	// node in struct tenstorrent_device.open_fds_list
 };
 
 struct chardev_private *get_tenstorrent_priv(struct file *f);
--- a/device.h
+++ b/device.h
@@ -38,7 +38,7 @@ struct tenstorrent_device {
 	struct tt_hwmon_context hwmon_context;
 	struct tt_attribute_data *attributes;
 
-	struct list_head open_fds;
+	struct list_head open_fds_list;	// List of struct chardev_private, linked through open_fds field
 };
 
 struct tenstorrent_device_class {
--- a/device.h
+++ b/device.h
@@ -37,6 +37,8 @@ struct tenstorrent_device {
 
 	struct tt_hwmon_context hwmon_context;
 	struct tt_attribute_data *attributes;
+
+	struct list_head open_fds;
 };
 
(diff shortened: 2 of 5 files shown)


joelsmithTT:
> What happens to mmapped DMA buffers when dma_free_coherent is called? The mapping still exists, what's backing it?

It looks like mmap increases the refcounts of the pages, so the buffer survives.

chardev_private.h
```diff
@@ -33,6 +33,8 @@ struct chardev_private {
 	struct list_head peer_mappings; // struct peer_resource_mapping.list
 
 	DECLARE_BITMAP(resource_lock, TENSTORRENT_RESOURCE_LOCK_COUNT);
+
+	struct list_head node;	// for struct tenstorrent_device::open_fds
```
alewycky-tenstorrent, line 37: Need a better name for "node": open_fds is okay and rename the head to open_fds_list?

device.h
```diff
@@ -37,6 +37,8 @@ struct tenstorrent_device {
 
 	struct tt_hwmon_context hwmon_context;
 	struct tt_attribute_data *attributes;
+
+	struct list_head open_fds;
```
alewycky-tenstorrent, line 41: Needs a comment saying what type is on this list (and which field is used for the list_head).

enumerate.c
```diff
@@ -116,6 +119,13 @@ static int tenstorrent_pci_probe(struct pci_dev *dev, const struct pci_device_id
 static void tenstorrent_pci_remove(struct pci_dev *dev)
 {
 	struct tenstorrent_device *tt_dev = pci_get_drvdata(dev);
+	struct chardev_private *priv, *tmp;
+
+	mutex_lock(&tt_dev->chardev_mutex);
```
alewycky-tenstorrent, line 124: tenstorrent_memory_cleanup also takes chardev_mutex but mutexes don't allow recursive locking. (https://docs.kernel.org/locking/mutex-design.html#semantics)

chardev_private.h
joelsmithTT: I renamed node to open_fd, and the list head from open_fds to open_fds_list.

device.h
joelsmithTT: please take another look

enumerate.c
joelsmithTT: I don't see where this happens.

alewycky-tenstorrent: Sorry that was in uncommitted code. There's still a problem. tenstorrent_pci_remove calls tenstorrent_memory_cleanup (holding tt_dev-scoped mutex). This can race against:
- close fd -> tt_cdev_release -> tenstorrent_memory_cleanup (no locking because it's called by the last owner)
- ioctl_allocate_dma_buf (only holds chardev_priv->mutex)
- ioctl_pin_pages & ioctl_unpin_pages (chardev_priv->mutex)
- ioctl_map_peer_bar (own chardev_priv->mutex and peer's chardev_priv->mutex)

Probably we should just take the chardev_priv mutex in tenstorrent_memory_cleanup. So far no code holds both mutexes at the same time.

PR #37, opened 2025-01-29
Fix `error: 'numeric_limits' is not a member of 'std'`

Artturin:
The build failed with that error in nix with gcc13

`make -C test`

Ref https://stackoverflow.com/questions/71296302/numeric-limits-is-not-a-member-of-std


alewycky-tenstorrent:
Thank you for fixing this!

PR #46, opened 2025-03-10
Added additional board types for new WH and BH cards

TTDRosen:
Added the board types for Galaxy UBB;p150a;p150b;p150c and p100a.

### DIFF
--- a/blackhole.c
+++ b/blackhole.c
@@ -126,6 +126,7 @@ static const struct blackhole_hwmon_attr bh_hwmon_attrs[] = {
 };
 
 // Prototypes for device_attribute show functions
+static ssize_t bh_show_card_type(struct device *dev, struct device_attribute *attr, char *buf);
 static ssize_t bh_show_attribute(struct device *dev, struct device_attribute *attr, char *buf);
 static ssize_t bh_show_card_serial(struct device *dev, struct device_attribute *attr, char *buf);
 static ssize_t bh_show_fw_ver(struct device *dev, struct device_attribute *attr, char *buf);
@@ -138,6 +139,7 @@ static const struct tt_attribute_data bh_attributes[] = {
 	{ __ATTR(tt_serial, S_IRUGO, bh_show_card_serial, NULL) },
 	{ __ATTR(tt_m3app_fw_ver, S_IRUGO, bh_show_fw_ver, NULL) },
 	{ __ATTR(tt_fw_bundle_ver, S_IRUGO, bh_show_fw_ver, NULL) },
+	{ __ATTR(tt_card_type,  S_IRUGO, bh_show_card_type, NULL) },
 	{ __ATTR_NULL }
 };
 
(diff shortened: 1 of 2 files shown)


blackhole.c
```diff
@@ -152,6 +154,44 @@ static const int bh_attributes_ids[] = {
 	-1,
 };
 
+static bool bh_board_type_to_name(u32 board_type, char **name) {
```
alewycky-tenstorrent, line 157: Please const char **name.

blackhole.c
```diff
@@ -152,6 +154,44 @@ static const int bh_attributes_ids[] = {
...
+			return false;
+	}
+
+	return true;
+}
+
+static ssize_t bh_show_card_type(struct device *dev, struct device_attribute *attr, char *buf) {
+	struct tenstorrent_device *tt_dev = dev_get_drvdata(dev);
+	struct blackhole_device *bh = tt_dev_to_bh_dev(tt_dev);
+	char *card_name;
```
alewycky-tenstorrent, line 174: Please const char *card_name.

blackhole.c
```diff
@@ -152,6 +154,44 @@ static const int bh_attributes_ids[] = {
...
+	u32 j;
+	u64 addr = 0;
+
+	if (!bh_board_type_to_name(tt_dev->pdev->subsystem_device, &card_name)) {
+		// Reading back invalid board type from pcie config space.
```
joelsmithTT, line 179: Is this a big enough problem that attempting NOC access would make things worse?

TTDRosen: My logic was that one more pci read during boot was unlikely to cause any significant problems. Even if there was some init problem and may help us in cases where someone has mocked the PCIe config space i.e. in a VM. But that's also maybe me overthinking the problem cause I'm so used to working on half working products 😅

blackhole.c
TTDRosen: Fixed

blackhole.c
TTDRosen: Fixed

blackhole.c
TTDRosen: I also noticed that my board_type from board_id logic was also wrong... fixed now

blackhole.c
```diff
@@ -152,6 +154,51 @@ static const int bh_attributes_ids[] = {
 	-1,
 };
 
+static bool bh_board_type_to_name(u32 board_type, const char **name) {
```
olofj, line 157: Is there a reason you're not returning the pointer and NULL when there's no match? Much simpler interface for the same functionality (and common kernel pattern).

TTDRosen: I wanted to set the board_type to "unknown" inside the `bh_board_type_to_name` function. Then returned bool cause there were only two conditions pass or fail. But I'm not concerned about swapping it for what I think you mean by the more standard

```
static const char *bh_board_type_to_name(u32 board_type, const char **name);
```

blackhole.c
```diff
@@ -152,6 +154,51 @@ static const int bh_attributes_ids[] = {
...
+	switch (board_type) {
+	case 0x36: *name = "p100"; break;
+	case 0x40: *name = "p150a"; break;
+	case 0x41: *name = "p150b"; break;
+	case 0x42: *name = "p150c"; break;
+	case 0x43: *name = "p100a"; break;
+	default:
+			*name = "unknown";
```
joelsmithTT, line 165: nit: spacing

hwmon.c
```diff
@@ -93,6 +93,11 @@ ssize_t tt_show_card_type(struct device *dev, struct device_attribute *attr, cha
 	case 0x7: card_name = "e75"; break;
 	case 0x14: card_name = "n300"; break;
 	case 0x18: card_name = "n150"; break;
+	case 0x35: card_name = "galaxy-wormhole"; break;
+	case 0x40: card_name = "p150a"; break;
+	case 0x41: card_name = "p150b"; break;
+	case 0x42: card_name = "p150c"; break;
+	case 0x43: card_name = "p100a"; break;
```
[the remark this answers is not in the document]
joelsmithTT: I think you can remove the p* cases

joelsmithTT:
Left a few more comments, but looks reasonable to me.
I'll defer to @alewycky-tenstorrent for final approval

alewycky-tenstorrent (approved):

PR #63, opened 2025-05-15
ioctl: Clarify uapi licence

shenki:
The ioctl header is expected to be used like any other linux uapi header, where an exception for derived works is made. This is called the 'syscall-note' in the Linux kernel tree:

 https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/LICENSES/exceptions/Linux-syscall-note


warthog9 (approved):
LGTM

joelsmithTT (approved):

PR #65, opened 2025-05-16
Updated driver API to version 2

DRvader:
Driver version 1.33 had a bug in the implementation of the tlb allocation API, which required an API breaking change to fix in 1.34. To reflect this breaking change I have updated the driver API version to 2


joelsmithTT (approved):
LGTM, @alewycky-tenstorrent FYI

PR #71, opened 2025-05-29
module: added alpine support

rekudyu:
Alpine linux uses akms instead of dkms, this PR make sure that the module compiles and installed on alpine linux as per https://github.com/tenstorrent/tt-metal/issues/18290 requirement.


rekudyu:
forgot to include CI, working on it.

PR #72, opened 2025-05-29
Fix up test/tlbs.cpp

joelsmithTT:
These tests were developed against preproduction hardware (unharvested, no hardware coordinate translation) and made invalid assumptions:

1. DRAM core at (0, 0) is valid to use.
2. Tensix coordinates read from the NOC node ID register match the untranslated NOC0 coordinate.

This commit fixes the tests by:

- using NOC_NODE_ID for untranslated hardware and NOC_NODE_ID_LOGICAL for translated hardware.
- using telemetry to determine a valid DRAM core to use.


test/tlbs.cpp
```diff
@@ -139,6 +141,106 @@ void fill_with_random_data(std::vector<uint32_t> &data)
...
+}
+
+bool blackhole_get_valid_dram_coordinates(const EnumeratedDevice &dev, uint32_t &x, uint32_t &y)
+{
+    // 1st DRAM core is at (x=17, y=12) in the translated coordinate scheme.
+    x = 17;
+    y = 12;
+
+    // If the device is not using translated coordinates, we need some logic
+    // to determine an appropriate DRAM core to use.
```
alewycky-tenstorrent, line 220: It didn't occur to me before, but if there's no translation, then there's no harvesting and you can just use 0-0.

joelsmithTT: That's a big simplification.

test/tlbs.cpp
```diff
@@ -7,6 +7,7 @@
 
 #include <algorithm>
 #include <array>
+#include <iostream>
```
alewycky-tenstorrent, line 10: Don't think this is needed.

PR #74, opened 2025-05-30
module: added akms(alpinelinux) support

rekudyu:
this pr adds support for akms ( alpine linux dkms alternate). As for CI I am having difficulty as alpine is not directly provided in the CI(maybe using doker) but I am working on it.


rekudyu:
> Tentatively like the direction this is heading, would love to see a ci/cd workflow that builds an actual package, take a look here to double check how that might work: https://github.com/warthog9/luwen/blob/main/.github/workflows/build-deb-ubuntu-22.04.yml

Okay I will be reopening new pull request with all the required changes and package automation.

PR #79, opened 2025-06-03
Allow driver to build when hwmon is disabled

shenki:
When the kernel has CONFIG_HWMON=n tt-kmd fails to link due to the missing devm_hwmon_device_register_with_info symbol.

In order to allow testing with eg. Ubuntu 'cloud' kernels, provide a stub for this case.

Without hwmon support the user will miss out on all of the hwmon sensors (and the driver might error out before finishing probing?), so this is not a well supported use case, but it does help for build testing.


joelsmithTT (approved):
LGTM

PR #85, opened 2025-06-11
BH: add tt_asic_id to bh_attributes

kikkiakki:
Expose Blackhole ASIC ID as a device attribute. Reuse the bh_show_card_serial function as ASIC ID is formatted the same way as board ID in ARC FW telemetry.

The plan is to eventually do this for WH too but we still have to expose ASIC ID on the WH FW side.


joelsmithTT (approved):

olofj:
Is there anything holding up merging this? It's been reviewed.

Issue #86, opened 2025-06-14
Feature request: ioctl for noc_write as part of chardev release


pcawleyTT:
It is common for host software to launch some code running on the device, and leave that code running on the device for as long as the host software is running (I'm thinking of e.g. tt-metal's fast dispatch kernels, which sit running on the device waiting for commands to appear in a command queue). If the host software is abnormally terminated (e.g. segfault or OOM killer), then the code on the device can continue running. It would be good if there was a mechanism for the device software to be informed that the host software is no longer running. If the host software shuts down cleanly, then the host software can do this itself, but it can't in the case of abnormal termination (a reasonable analogy is robust futexes: most futex operations are entirely in user-space, but the robust futex feature is there in kernel-space to allow cleanup in case of abnormal termination).

One possible approach would be a new ioctl which accepts a structure like the following, and copies it to an identical structure within `chardev_private`:
```c
struct ioctl_noc_write_on_release {
  u8 enabled;
  u8 x;
  u8 y;
  u8 noc;
  u32 data;
  u64 addr;
};
```
Then in `tt_cdev_release`, before the `tenstorrent_memory_cleanup` call, something like:
```c
if (enabled && is_valid_noc_address(x, y, noc)) {
  noc_write32(device, x, y, addr, data, noc);
}
```
Implementation of `is_valid_noc_address` left as an exercise to the reader; a simple implementation is `x < 64 && y < 64 && noc < 2`, but you can be fancier if you want (query the NIU to get the NoC translation tables, apply a translation if translation is enabled, then query the NIU details to get the NoC width and height, and check the result is within bounds).

joelsmithTT:
Seems reasonable, some initial questions:

* Per file descriptor, should the API support 1 or N such writes?
* Is multicast a requirement?

pcawleyTT:
I'd be content with 1/unicast; that's enough to notify one on-device riscv, and it can then do N/multicast as part of its cleanup if it needs to.

PR #89, opened 2025-06-16
Add ioctl for robust cleanup via NOC write on release

joelsmithTT:
Resolves https://github.com/tenstorrent/tt-kmd/issues/86

This PR introduces a mechanism to allow device-side software to perform cleanup when the corresponding host process terminates abnormally.

**Technique**
A new ioctl, `TENSTORRENT_IOCTL_SET_NOC_CLEANUP`, is added.  It allows a process to register a single 32-bit NOC write operation (target coordinates, address, and data) with the kernel.  This operation is tied to the file descriptor and is automatically executed by the driver when the file descriptor is closed for any reason.

**Changes**
* **New IOCTL API Pattern**: This `ioctl` intentionally deviates from the legacy in/out struct pattern.  It passes a single data structure directly, an approach inspired by the VFIO API.  This new pattern, which includes `argsz` and `flags` fields for versioning/compatibility checks, is intended to be the standard for future ioctl additions.  Existing ioctls are left untouched for backward compatibility.
* `ioctl.h`: Defines the new `TENSTORRENT_IOCTL_SET_NOC_CLEANUP` and its associated struct.
* `chardev.c`:
  * Implements the new ioctl, validating user input and storing the action in the per-FD `chardev_private` data.
  * Updates `tt_cdev_release()` to execute the registered write action upon file close.
* `device.h`, `wormhole.c`, `blackhole.c`: A `noc_write32` function pointer has been added to the `tenstorrent_device_class` to abstract the device-specific write logic.  This feature is implemented for Wormhole and Blackhole.
* `test/`: A new test suite (`release.cpp`) is added to validate the feature.  It confirms that the write is performed correctly on `close()` and that the action can be disabled.

### DIFF
--- a/test/Makefile
+++ b/test/Makefile
@@ -7,7 +7,7 @@ PROG := ttkmd_test
 
 TEST_SOURCES := get_driver_info.cpp get_device_info.cpp query_mappings.cpp \
 	dma_buf.cpp pin_pages.cpp config_space.cpp lock.cpp hwmon.cpp map_peer_bar.cpp \
-	ioctl_overrun.cpp ioctl_zeroing.cpp tlbs.cpp
+	ioctl_overrun.cpp ioctl_zeroing.cpp tlbs.cpp release.cpp
 
 CORE_SOURCES := enumeration.cpp util.cpp devfd.cpp main.cpp test_failure.cpp
 SOURCES := $(CORE_SOURCES) $(TEST_SOURCES)
--- a/test/ioctl.h
+++ b/test/ioctl.h
@@ -25,6 +25,7 @@
 #define TENSTORRENT_IOCTL_ALLOCATE_TLB		_IO(TENSTORRENT_IOCTL_MAGIC, 11)
 #define TENSTORRENT_IOCTL_FREE_TLB		_IO(TENSTORRENT_IOCTL_MAGIC, 12)
 #define TENSTORRENT_IOCTL_CONFIGURE_TLB		_IO(TENSTORRENT_IOCTL_MAGIC, 13)
+#define TENSTORRENT_IOCTL_SET_NOC_CLEANUP		_IO(TENSTORRENT_IOCTL_MAGIC, 14)
 
 // For tenstorrent_mapping.mapping_id. These are not array indices.
 #define TENSTORRENT_MAPPING_UNUSED		0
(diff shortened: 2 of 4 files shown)


chardev.c
```diff
@@ -412,6 +462,19 @@ static int tt_cdev_release(struct inode *inode, struct file *file)
 	struct tenstorrent_device *tt_dev = priv->device;
 	unsigned int bitpos;
 
+	if (priv->write_on_release.enabled) {
+		if (tt_dev->dev_class && tt_dev->dev_class->noc_write32) {
```
pcawleyTT, line 466: This is also checked in `ioctl_noc_write_on_release`. No particular harm in checking it in two places, though I find it odd that the code here checks for `tt_dev->dev_class != NULL`, whereas the code in `ioctl_noc_write_on_release` is only checking `tt_dev->dev_class->noc_write32 != NULL`. If both places are going to check it, might as well check the exact same thing in both.

ioctl.h
```diff
@@ -293,4 +296,34 @@ struct tenstorrent_configure_tlb {
...
+ * @y: Y coordinate of the NOC tile to write to.
+ * @noc: NOC ID to write to; must be 0 or 1.
+ * @data: Data to write to the NOC tile.
+ * @addr: NOC address to write to; must be 4-byte aligned.
+ */
+struct tenstorrent_noc_write_on_release {
+	__u32 argsz;
+	__u32 flags;
+	__u8 enabled;
+	__u8 x;
```
shenki, line 322: We spoke in person about making this generic to allow more than one write. I guess you can make this an array of x,y,noc,data,addr later without breaking the V1 API?

Adding a count field after enabled that must be 1 for the first version?

shenki (approved):
Minor comments.

ioctl.h
pcawleyTT: For extra context, my use-case involves a collection of cleanup actions: some NOC writes, some ethernet writes, and some sending of ARC messages (once you combine these actions, you get nice things like "on n300 boards, the clock speed of both the local ASIC and the remote ASIC is restored to idle speed as part of cleanup"). If it were _just_ NOC writes, I could see the argument for having kmd do all the writes, but once there are other types of action, it makes sense to have some on-device code respond to a NOC write and do the other types of cleanup action. Once you've got that, it's fairly easy to have that on-device code respond to just one NOC write and perform all the other NOC writes, meaning kmd only needs to do one NOC write.

blackhole.c
```diff
@@ -199,39 +199,43 @@ static int blackhole_configure_tlb_4G(struct blackhole_device *bh, int tlb,
 	return 0;
 }
 
-static u8 __iomem *bh_configure_kernel_tlb(struct blackhole_device *bh, u32 x, u32 y, u64 addr) {
+static u8 __iomem *bh_configure_kernel_tlb(struct blackhole_device *bh, u32 x, u32 y, u64 addr, int noc)
```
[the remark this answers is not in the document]
joelsmithTT: X and Y are the location on _a_ NOC of a tile.  The address is within the tile.  Depending on the HW translation scheme, the NOC0 and NOC1 coordinates of a tile may or may not be identical.

From coarsest to finest addressing parameter, it goes noc, xy, addr, so from this perspective it makes sense for noc to be first.  I'm not opposed to a patch making it come before x, but I am not going to do that for this work.

ioctl.h
```diff
@@ -293,4 +296,34 @@ struct tenstorrent_configure_tlb {
 	struct tenstorrent_configure_tlb_out out;
 };
 
+/**
+ * TENSTORRENT_IOCTL_NOC_WRITE_ON_RELEASE - Register a cleanup action
```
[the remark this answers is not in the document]
joelsmithTT: Good point.  I am inclined toward `TENSTORRENT_IOCTL_SET_NOC_CLEANUP`

ioctl.h
```diff
@@ -25,6 +25,9 @@
 #define TENSTORRENT_IOCTL_ALLOCATE_TLB		_IO(TENSTORRENT_IOCTL_MAGIC, 11)
 #define TENSTORRENT_IOCTL_FREE_TLB		_IO(TENSTORRENT_IOCTL_MAGIC, 12)
 #define TENSTORRENT_IOCTL_CONFIGURE_TLB		_IO(TENSTORRENT_IOCTL_MAGIC, 13)
+#define TENSTORRENT_IOCTL_NOC_WRITE_ON_RELEASE \
+	_IOWR(TENSTORRENT_IOCTL_MAGIC, 14,     \
```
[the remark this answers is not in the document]
alewycky-tenstorrent: Agree with _IO. Baking a data size into _IOWR gets confusing when we extend the ioctl data.

ioctl.h
```diff
@@ -293,4 +294,34 @@ struct tenstorrent_configure_tlb {
...
+ * @addr: NOC address to write to; must be 4-byte aligned.
+ */
+struct tenstorrent_set_noc_cleanup {
+	__u32 argsz;
+	__u32 flags;
+	__u8 enabled;
+	__u8 x;
+	__u8 y;
+	__u8 noc;
+	__u32 data;
```
alewycky-tenstorrent, line 323: Based on our slack discussion, widen data to __u64 for easy future extension, but only write the low 32 bits now.

alewycky-tenstorrent (approved):

PR #92, opened 2025-06-19
Remove community-issue-tagging.yml

joelsmithTT:
This file causes a failure email to be generated on events like GitHub issue creation.


joelsmithTT:
@warthog9 I don't understand the intent of this file, can we either fix it up (so it does whatever is intended without emailing out errors) or remove it until a fix is ready (as this PR does)?  Thanks

joelsmithTT:
Thanks, I will give that a try.

PR #94, opened 2025-07-03
Add nix flake

RossComputerGuy:
Since [nixpkgs itself has tt-kmd](https://github.com/NixOS/nixpkgs/pull/421385), let's add a Nix Flake here. I've gone ahead and add a job to the CI for it. I could integrate VM testing to at least test if the driver loads. Aside from that, this should work.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,6 +7,17 @@ on:
 
 jobs:
 
+  build-nix:
+    strategy:
+      matrix:
+        os: [ubuntu-24.04, ubuntu-22.04]
+    runs-on: ${{ matrix.os }}
+    steps:
+    - uses: actions/checkout@v4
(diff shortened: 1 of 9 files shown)


RossComputerGuy:
> Could you explain the advantage users get from having this here versus using what's in the nixpkgs?

This allows for being able to develop the driver via `nix develop` in this repository. Nixpkgs only packages major versions of packages unless an exception is warranted. Having a Flake here makes it possible to work on the module and use an unstable version of it.

> if we merge this, would you be willing to maintain it going forward?

Yes 

> Nix seems like it requires specialized knowledge (i.e. how to generate the hashes I see in `flake.lock`) that I don't have.

`nix flake update` / `nix flake lock`

RossComputerGuy:
I've gone ahead and done that.

PR #97, opened 2025-07-09
fix: Validate CSM access attempts

joelsmithTT:
When a chip is in a bad state, an ioread32 operation to fetch a pointer from the ARC tile can return all ones. If the driver subsequently uses this garbage value as an address for a noc_read32/write32 operation, it will access an invalid memory location and malfunction.

This commit hardens the ARC message-passing logic by introducing csm_read32 and csm_write32 helper functions. These wrappers validate that any address is within the legal CSM range before executing the hardware access. The message-passing functions are refactored to use these safe wrappers, which now propagate failures and prevent the use of bogus addresses read from hardware in a fault state.

Additionally, this change fixes a latent bug in the is_address_within_csm boundary check. The original logic was incorrect and would have permitted a 32-bit read to start at the final byte of the CSM. The check is corrected to ensure the entire 32-bit access remains within the valid range.


blackhole.c
```diff
@@ -253,6 +253,32 @@ static void noc_write32(struct blackhole_device *bh, u32 x, u32 y, u64 addr, u32
 	mutex_unlock(&bh->kernel_tlb_mutex);
 }
 
+#define ARC_CSM_BASE 0x10000000
+#define ARC_CSM_SIZE (1 << 19)
+static bool is_address_within_csm(u64 addr)
```
alewycky-tenstorrent, line 258: I think this should take the access size. There are a couple existing callers that could just pass 1 or something.

joelsmithTT: That's what I wrote first (I renamed it to `is_range_within_csm(u64, size_t)`, but then reverted it).  Everywhere I call it, I was passing sizeof(u32).  By 1, you mean "one u32"?

alewycky-tenstorrent (approved):

Issue #82, opened 2025-06-04
KMD should only allocate exact sizes


DRvader:
The current driver behaviour is to allocate the smallest tlb that fits the requested size. This is an issue because you need to know the size of the underlying tlb in order to correctly calculate the address mask. I work around this by only requesting tlbs in exact sizes, but to avoid confusion I think it would be best to either return the size of the backing tlb or to error out if the requested size does not exactly match a tlb_kind.

joelsmithTT:
My preference is the latter (`ALLOCATE_TLB` fails if `tenstorrent_allocate_tlb_in.size` does not match one of the hardware's window sizes) because it's simpler than returning the actual size.  At this stage it is unlikely to break anybody.

In retrospect, the driver behavior of silently rounding up seems like a bug.

PR #102, opened 2025-07-15
fix(tlb): Require exact size on TLB allocation

joelsmithTT:
The previous TLB allocation logic would accept any size and silently round it up to the next largest available TLB window size. This could mask bugs in userspace code, which should always request a specific, valid TLB window size.

This commit enforces a stricter policy by changing the lookup condition from a less-than-or-equal check to an exact-match check. The driver will now only allocate a TLB window if the requested size precisely matches a valid hardware window size.


joelsmithTT:
Fixes https://github.com/tenstorrent/tt-kmd/issues/82

PR #108, opened 2025-07-22
Asahi

fede2cr:
This pull request adds support for Asahi.

It creates a SPEC file that build an rpm under github actions. The github action must use a local runner, as github doesn't have native arm64 runners and also doesn't have containers or vms with Asahi.


fede2cr:
It's one of the steps for tenstorrent/tt-metal#18296.

joelsmithTT:
Thank you for the contribution.  I am closing this PR without merging because the target Asahi platform lacks the PCIe support that is a fundamental requirement for this driver to function.  I will note on the bounty ticket that this work was completed as requested.

PR #129, opened 2025-08-21
Improve Wormhole reset robustness

joelsmithTT:
1. Set the reset marker prior to triggering a reset.
2. Enter a polling loop to wait out the auto-reset watchdog if the device appears to be hung.

### DIFF
--- a/wormhole.c
+++ b/wormhole.c
@@ -319,25 +319,50 @@ static void update_device_index(struct wormhole_device *wh_dev) {
 static bool wormhole_reset(struct tenstorrent_device *tt_dev, u32 reset_flag)
 {
 	struct pci_dev *pdev = tt_dev->pdev;
+	struct wormhole_device *wh_dev = tt_dev_to_wh_dev(tt_dev);
 	u16 reset_arg = (reset_flag == TENSTORRENT_RESET_DEVICE_ASIC_DMC_RESET) ? 3 : 0;
+	bool responsive;
 
-	// Step 1: Secondary bus reset.
+	// 1. Attempt a secondary bus reset.
 	pcie_hot_reset_and_restore_state(pdev);
 
-	// Step 2: Check if the device is functioning enough to respond to a message.
-	if (!grayskull_send_arc_fw_message_with_args(reset_unit_regs(tt_dev_to_wh_dev(tt_dev)), WH_FW_MSG_NOP,
-						     reset_arg, 0, 10000, NULL)) {
-		dev_warn(&tt_dev->pdev->dev, "Couldn't communicate with firmware; NOC is likely hung.\n");
-		return false;
+	// 2. See if the device is responsive.
+	responsive = grayskull_send_arc_fw_message(reset_unit_regs(wh_dev), WH_FW_MSG_NOP, 1000, NULL);
+
+	// 3. If not responsive, wait for the watchdog.
+	if (!responsive) {
+		ktime_t end_time;
+
+		// Give up immediately if the watchdog is disabled.
+		if (auto_reset_timeout == 0) {
+			dev_err(&pdev->dev, "Watchdog is disabled and device is unresponsive, cannot reset.\n");
+			return false;
+		}
+
+		end_time = ktime_add_ms(ktime_get(), (auto_reset_timeout * 1000) + 500);
+		while (ktime_before(ktime_get(), end_time)) {
+			pcie_hot_reset_and_restore_state(pdev);
+
+			responsive = grayskull_send_arc_fw_message(reset_unit_regs(wh_dev), WH_FW_MSG_NOP, 1000, NULL);
+			if (responsive)
+				break;
+
+			if (msleep_interruptible(1000))
+				return false;
+		}
 	}
 
-	// Step 3: trigger the reset.
-	grayskull_send_arc_fw_message_with_args(reset_unit_regs(tt_dev_to_wh_dev(tt_dev)), WH_FW_MSG_TRIGGER_RESET,
-						reset_arg, 0,
-						0,
-						NULL);
+	// 4. If the device is responsive, finalize the reset.
+	if (responsive) {
+		set_reset_marker(pdev);
+		grayskull_send_arc_fw_message_with_args(reset_unit_regs(wh_dev), WH_FW_MSG_TRIGGER_RESET, reset_arg, 0,
+							0, NULL);
+		return true; // Assumes the reset was successful.
+	}
 
-	return true; // Assumes the reset was successful.
+	// If we've reached this point, the device is still unresponsive.
+	dev_err(&pdev->dev, "Device is unresponsive, cannot reset.\n");
+	return false;
 }
 
 static int telemetry_probe(struct tenstorrent_device *tt_dev)
--- a/grayskull.c
+++ b/grayskull.c
@@ -126,6 +126,9 @@ bool grayskull_send_arc_fw_message_with_args(u8 __iomem* reset_unit_regs,
 	arc_misc_cntl = ioread32(arc_misc_cntl_reg);
 	iowrite32(arc_misc_cntl | ARC_MISC_CNTL_IRQ0_MASK, arc_misc_cntl_reg);
 
+	if (timeout_us == 0)
+		return false;
+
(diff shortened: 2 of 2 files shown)


wormhole.c
```diff
@@ -319,23 +319,44 @@ static void update_device_index(struct wormhole_device *wh_dev) {
...
 	}
 
-	// Step 3: trigger the reset.
-	grayskull_send_arc_fw_message_with_args(reset_unit_regs(tt_dev_to_wh_dev(tt_dev)), WH_FW_MSG_TRIGGER_RESET,
-						reset_arg, 0,
-						0,
-						NULL);
+	end_time = ktime_add_ms(ktime_get(), (auto_reset_timeout * 1000) + 500);
+	while (ktime_before(ktime_get(), end_time)) {
+		pcie_hot_reset_and_restore_state(pdev);
```
joelsmithTT, line 343: This repeated `pcie_hot_reset_and_restore_state` seems to be necessary in the hard hang case.  Without it, I get e.g.

```
Aug 21 21:09:28 ubuntu kernel: [100111.210705] tenstorrent: Skipping message 00000011 due to FW not running.
Aug 21 21:09:29 ubuntu kernel: [100112.234690] tenstorrent: Skipping message 00000011 due to FW not running.
Aug 21 21:09:30 ubuntu kernel: [100113.258680] tenstorrent 0000:02:00.0: Device is unresponsive, cannot reset.
```

wormhole.c
```diff
@@ -319,23 +319,44 @@ static void update_device_index(struct wormhole_device *wh_dev) {
...
 	pcie_hot_reset_and_restore_state(pdev);
 
-	// Step 2: Check if the device is functioning enough to respond to a message.
-	if (!grayskull_send_arc_fw_message_with_args(reset_unit_regs(tt_dev_to_wh_dev(tt_dev)), WH_FW_MSG_NOP,
-						     reset_arg, 0, 10000, NULL)) {
-		dev_warn(&tt_dev->pdev->dev, "Couldn't communicate with firmware; NOC is likely hung.\n");
+	// 2. See if the device is responsive.
+	responsive = grayskull_send_arc_fw_message_with_args(reset_unit_regs(wh_dev), WH_FW_MSG_NOP, 0, 0, 1000, NULL);
+	if (responsive)
+		goto finalize_reset;
```
alewycky-tenstorrent, line 333: I don't think the gotos are justified. This can be an ordinary if and the later goto can be a break.

wormhole.c
nxuTT: I think it makes sense, in auto reset case, after chip config space got cleared, kmd needs to restore config space to set ARC message. And pcie_hot_reset_and_restore_state needs to be done after auto reset has completed

wormhole.c
```diff
@@ -319,23 +319,44 @@ static void update_device_index(struct wormhole_device *wh_dev) {
...
-	// Step 3: trigger the reset.
-	grayskull_send_arc_fw_message_with_args(reset_unit_regs(tt_dev_to_wh_dev(tt_dev)), WH_FW_MSG_TRIGGER_RESET,
-						reset_arg, 0,
-						0,
-						NULL);
+	end_time = ktime_add_ms(ktime_get(), (auto_reset_timeout * 1000) + 500);
+	while (ktime_before(ktime_get(), end_time)) {
+		pcie_hot_reset_and_restore_state(pdev);
+
+		responsive = grayskull_send_arc_fw_message_with_args(reset_unit_regs(wh_dev), WH_FW_MSG_NOP, 0, 0, 1000, NULL);
```
alewycky-tenstorrent, line 345: Please use grayskull_send_arc_fw_message when both args are 0.

wormhole.c
```diff
@@ -319,23 +319,44 @@ static void update_device_index(struct wormhole_device *wh_dev) {
...
+
+		if (responsive)
+			goto finalize_reset;
+
+		msleep(1000);
```
alewycky-tenstorrent, line 350: How about msleep_interruptible? Can we safely give up here if the user wants out?

wormhole.c
```diff
@@ -319,23 +319,44 @@ static void update_device_index(struct wormhole_device *wh_dev) {
...
+	}
+
+	dev_err(&pdev->dev, "Device is unresponsive, cannot reset.\n");
+	return false;
+
+	// 4. Set the reset marker and trigger the reset.
+finalize_reset:
+	set_reset_marker(pdev);
+	grayskull_send_arc_fw_message_with_args(reset_unit_regs(wh_dev), WH_FW_MSG_TRIGGER_RESET, reset_arg, 0, 0, NULL);
```
alewycky-tenstorrent, line 359: When you send the reset message, you can't wait for the FW to acknowledge it. The device may be instantly reset, possibly before it sets the scratch response, possibly after but before you read it.

I suggest splitting grayskull_send_arc_fw_message_with_args to create a special case function that doesn't check for results at all.

joelsmithTT: Would you accept U32_MAX as a magic timeout for don't check completion?

joelsmithTT: Ok.  I checked all the call sites for both send_with_args and without; only here is using a timeout of 0 to indicate early return of false (no completion checking).

alewycky-tenstorrent (approved):

wormhole.c
```diff
@@ -319,25 +319,50 @@ static void update_device_index(struct wormhole_device *wh_dev) {
...
 	}
 
-	// Step 3: trigger the reset.
-	grayskull_send_arc_fw_message_with_args(reset_unit_regs(tt_dev_to_wh_dev(tt_dev)), WH_FW_MSG_TRIGGER_RESET,
-						reset_arg, 0,
-						0,
-						NULL);
+	// 4. If the device is responsive, finalize the reset.
+	if (responsive) {
+		set_reset_marker(pdev);
```
nbuncicTT, line 357: Just for my understanding: We use `set_reset_marker` to set the a bit in the config space prior to sending a reset signal here even though the board has been reset by the watchdog, because we want to trigger our own reset after the watchdog reset? And then later, via IOCTL, we check if this bit was cleared with our reset?

wormhole.c
```diff
@@ -319,25 +319,50 @@ static void update_device_index(struct wormhole_device *wh_dev) {
...
+	// 1. Attempt a secondary bus reset.
 	pcie_hot_reset_and_restore_state(pdev);
 
-	// Step 2: Check if the device is functioning enough to respond to a message.
-	if (!grayskull_send_arc_fw_message_with_args(reset_unit_regs(tt_dev_to_wh_dev(tt_dev)), WH_FW_MSG_NOP,
-						     reset_arg, 0, 10000, NULL)) {
-		dev_warn(&tt_dev->pdev->dev, "Couldn't communicate with firmware; NOC is likely hung.\n");
-		return false;
+	// 2. See if the device is responsive.
+	responsive = grayskull_send_arc_fw_message(reset_unit_regs(wh_dev), WH_FW_MSG_NOP, 1000, NULL);
```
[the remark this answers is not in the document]
joelsmithTT: See [this comment](https://github.com/tenstorrent/tt-kmd/blob/57045432e07789f000a50f49d7ac42c0d22aa6d6/grayskull.c#L4-L15) for some context.  But yes, s/grayskull/wormhole should happen.

wormhole.c
```diff
@@ -319,25 +319,50 @@ static void update_device_index(struct wormhole_device *wh_dev) {
...
+
+	// 3. If not responsive, wait for the watchdog.
+	if (!responsive) {
+		ktime_t end_time;
+
+		// Give up immediately if the watchdog is disabled.
+		if (auto_reset_timeout == 0) {
```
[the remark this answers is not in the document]
joelsmithTT: I agree that your naming suggestion makes the code more readable, although this value is a module parameter:

```
sudo modprobe tenstorrent auto_reset_timeout=0  # 0 disables the watchdog
```

The tradeoff is clarity of name vs silent breakage (`sudo modprobe tenstorrent does_not_exist=1` does not generate any error, although it is logged: `tenstorrent: unknown parameter 'does_not_exist' ignored`).  Although this is easier to change than the ioctl API, it runs the risk of surprises for anyone (probably syseng) who is overriding it.

PR #130, opened 2025-08-28
CI: Use explicit version string for releases

joelsmithTT:
Refactors the release workflow to be driven by an exact version string provided at dispatch time, rather than a version bump level (patch, minor, major).

Previously, the release process was triggered by selecting a bump level, which made creating pre-releases (e.g. 2.4.0-rc1) difficult.

The entirety of this patch was written by AI and manually tested.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,6 +3,11 @@ run-name: Create Release
 
 on:
   workflow_dispatch:
+    inputs:
+      version_string:
+        description: 'The exact version to release (e.g., 2.4.0 or 2.4.0-rc1)'
+        required: true
+        type: string
   workflow_call:
 
 env:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -251,6 +232,7 @@ jobs:
   # Trigger Building Everything
   ###
   build_all_depends:
+    if: github.repository_owner == 'tenstorrent'
```
joelsmithTT, line 235: This and the `if: always()` below were added when I tested this on a fork.  The goal was to skip package building (which fails on my non-tenstorrent fork due to lack of signing keys).  I also had to comment out a `files:` section when I tested.

alewycky-tenstorrent:
Pre-existing issues:
- tagrelease creates a lightweight tag rather than annotated tag
- [bump2version](https://pypi.org/project/bump2version/) [advises users to switch](https://github.com/c4urself/bump2version?tab=readme-ov-file#bump2version) to [bump-my-version](https://pypi.org/project/bump-my-version/)

I don't know how, but please ensure this workflow can only be activated by maintainers.

PR #143, opened 2025-10-15
Refcount bar_mapping structs

joelsmithTT:
Bug:
- BAR mappings had no .open callback or refcounting
- When a process forked after mmap'ing a BAR, both parent and child VMAs pointed to the same struct bar_mapping
- First process to exit called bar_vma_close() which freed the mapping struct
- Second process to exit triggered use-after-free

Fix:
- Added refcount_t refs to struct bar_mapping
- Implemented .open callback to increment refcount on fork
- Modified .close to only free when refcount reaches zero
- Added refcount display to debugfs mappings file for diagnostics

### DIFF
--- a/chardev_private.h
+++ b/chardev_private.h
@@ -8,6 +8,7 @@
 #include <linux/mutex.h>
 #include <linux/hashtable.h>
 #include <linux/sched.h>
+#include <linux/refcount.h>
 
 #include "ioctl.h"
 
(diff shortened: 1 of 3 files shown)


chardev_private.h
```diff
@@ -22,6 +22,7 @@ struct bar_mapping {
 	u64 size;
 	unsigned int bar_index;
 	enum bar_mapping_type type;
+	atomic_t refs;
```
alewycky-tenstorrent: Does refcount_t work for this purpose? I don't think it's really much different, but it adds overflow checks.

alewycky-tenstorrent (approved):

Issue #131, opened 2025-09-03
Zero the stride register in CONFIGURE_TLB path on Blackhole


joelsmithTT:
The first 32 2M TLB windows in Blackhole have additional configuration registers for advanced multicast targeting.  The driver does not expose this functionality.  The driver should zero these registers when it programs one of these windows for normal use.  The rationale is that if userspace takes advantage of this hardware feature, the state could persist beyond the lifetime of the application.

PR #151, opened 2025-10-21
blackhole: Zero strided TLB registers in configure path

joelsmithTT:
Blackhole's first 32 TLB windows have additional strided registers (4 bytes each) that control non-rectangular multicast patterns including x_keep/x_skip, y_keep/y_skip, and exclusion zones. These registers are located at offset 2520 (210 TLBs × 12 bytes) from TLB_REGS_START.

Since the TENSTORRENT_IOCTL_CONFIGURE_TLB interface does not expose these strided configuration parameters through the tenstorrent_noc_tlb_config structure, we must explicitly zero them when configuring TLBs 0-31 to prevent stale values from causing unintended multicast behavior.

Without this change, the strided registers could retain values from previous configurations, potentially causing multicast operations to skip tiles or exclude regions that the user did not intend.

This change only affects Blackhole, as the strided registers are part of its TLB configuration interface. The ioctl API intentionally does not support non-rectangular multicast configuration at this time.

### DIFF
--- a/blackhole.c
+++ b/blackhole.c
@@ -27,9 +27,14 @@
 #define TLB_4G_WINDOW_MASK (TLB_4G_WINDOW_SIZE - 1)
 
 #define TLB_REG_SIZE 12	// Same for 2M and 4G
+#define TLB_TOTAL_WINDOW_COUNT (TLB_2M_WINDOW_COUNT + TLB_4G_WINDOW_COUNT)
 
 #define TLB_REGS_START 0x1FC00000   // BAR0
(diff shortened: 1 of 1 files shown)


joelsmithTT:
Addresses https://github.com/tenstorrent/tt-kmd/issues/131

joelsmithTT:
For posterity - relevant documentation: https://github.com/tenstorrent/tt-isa-documentation/blob/main/BlackholeA0/PCIExpressTile/HostToDeviceTLBs.md

PR #158, opened 2025-10-27
CI: Use explicit version string for releases

joelsmithTT:
Refactors the release workflow to be driven by an exact version string provided at dispatch time.

Mostly AI-written.


.github/workflows/release.yml
```diff
@@ -58,7 +65,7 @@ jobs:
       git_hash: ${{ steps.new-git-hash.outputs.git_hash }}
       package_name: ${{ steps.packagename.outputs.package_name }}
       package_version: ${{ steps.packagever.outputs.package_version }}
-      package_version_new: ${{ steps.new-package-version.outputs.package_version_new }}
+      package_version_new: ${{ github.event.inputs.version_string }}
```
alewycky-tenstorrent: I think this can just be inputs.version_string. (https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#providing-inputs)

(And similar below.)

alewycky-tenstorrent (approved):

PR #160, opened 2025-10-31
Make struct padding explicit in ioctl.h

### LINKED ISSUE #107: Refactor: Make padding in ioctl.h structs explicit
Running `pahole` on the structures defined in ioctl.h reveals that several of them contain implicit padding inserted by the compiler for alignment purposes.  For example, pahole shows that tenstorrent_configure_tlb_in has a 4-byte hole:
```
struct tenstorrent_configure_tlb_in {
	__u32                      id;                   /*     0     4 */

	/* XXX 4 bytes hole, try to pack */

	struct tenstorrent_noc_tlb_config config;        /*     8    32 */

	/* size: 40, cachelines: 1, members: 2 */
	/* sum members: 36, holes: 1, sum holes: 4 */
	/* last cacheline: 40 bytes */
};
```

To make the user-kernel ABI more robust, make such padding explicit.

`struct tenstorrent_get_device_info_out`: tail padding of 2 bytes
`struct tenstorrent_lock_ctl_in`: tail padding of 3 bytes
`struct tenstorrent_map_peer_bar`: internal hole of 4 bytes
`struct tenstorrent_configure_tlb_in`: internal hole of 4 bytes

joelsmithTT:
Add explicit reserved fields to make compiler-inserted padding visible.

Affected structures:
- tenstorrent_get_device_info_out: added __u16 reserved (2-byte tail)
- tenstorrent_lock_ctl_in: added __u8 reserved[3] (3-byte tail)
- tenstorrent_lock_ctl_out: added __u8 reserved[3] (3-byte tail)
- tenstorrent_map_peer_bar: added __u32 reserved (4-byte internal hole)
- tenstorrent_configure_tlb_in: added __u32 reserved (4-byte internal hole)

All struct sizes and field offsets remain unchanged, preserving ABI compatibility. Verified with pahole and struct layout tests.

Fixes #107

### DIFF
--- a/test/ioctl.h
+++ b/test/ioctl.h
@@ -54,6 +54,7 @@ struct tenstorrent_get_device_info_out {
 	__u16 bus_dev_fn;	// [0:2] function, [3:7] device, [8:15] bus
 	__u16 max_dma_buf_size_log2;	// Since 1.0
 	__u16 pci_domain;		// Since 1.23
+	__u16 reserved;
 };
 
 struct tenstorrent_get_device_info {
@@ -136,10 +137,16 @@ struct tenstorrent_get_driver_info {
 	struct tenstorrent_get_driver_info_out out;
 };
 
-// tenstorrent_reset_device_in.flags
+// legacy tenstorrent_reset_device_in.flags
 #define TENSTORRENT_RESET_DEVICE_RESTORE_STATE 0
 #define TENSTORRENT_RESET_DEVICE_RESET_PCIE_LINK 1
 #define TENSTORRENT_RESET_DEVICE_CONFIG_WRITE 2
+
+// tenstorrent_reset_device_in.flags
+#define TENSTORRENT_RESET_DEVICE_USER_RESET 3
+#define TENSTORRENT_RESET_DEVICE_ASIC_RESET 4
+#define TENSTORRENT_RESET_DEVICE_ASIC_DMC_RESET 5
+#define TENSTORRENT_RESET_DEVICE_POST_RESET 6
 
 struct tenstorrent_reset_device_in {
 	__u32 output_size_bytes;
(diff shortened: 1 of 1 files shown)


joelsmithTT:
test/ioctl.h is deliberately left unchanged.

The reserved field in tenstorrent_map_peer_bar is placed between the in and out members rather than inside tenstorrent_map_peer_bar_in to avoid changing the _in struct size.  Only the overrun test uses _in standalone (which is otherwise useless since you can't retrieve the output), but expanding _in would technically break that compatibility.  Open to feedback on whether we should just expand _in and update the test instead, or keep this approach.

alewycky-tenstorrent:
How did you run pahole? Can we create a test that runs it?

test/pahole_check.sh
```diff
@@ -0,0 +1,123 @@
...
+        if [ -n "$OUTPUT" ]; then
+            echo "struct $struct:"
+            echo "$OUTPUT"
+            echo ""
+        fi
+    done
+fi
+
+# Cleanup
+rm -f "$SCRIPT_DIR/pahole_dummy.o" "$SCRIPT_DIR/pahole_dummy.c"
```
alewycky-tenstorrent: Use trap EXIT to clean up temporary files.

alewycky-tenstorrent:
> test/ioctl.h is deliberately left unchanged.
> 
> The reserved field in tenstorrent_map_peer_bar is placed between the in and out members rather than inside tenstorrent_map_peer_bar_in to avoid changing the _in struct size. Only the overrun test uses _in standalone (which is otherwise useless since you can't retrieve the output), but expanding _in would technically break that compatibility. Open to feedback on whether we should just expand _in and update the test instead, or keep this approach.

We can't really maintain separate versions of the header file. test should already be using ../ioctl.h.

ioctls only need to be compatible with callers that pass the whole structure. Putting the padding into tenstorrent_map_peer_bar_in means that old ttkmd_test binaries would fail overrun on new kmd. As long as they don't fail any functional test, that's fine.

PR #161, opened 2025-10-31
ci: retry apt-get update until lock is released

joelsmithTT:
Fixes intermittent CI failures caused by background apt processes holding the dpkg lock on self-hosted runners.

Example: https://github.com/tenstorrent/tt-kmd/actions/runs/18959560228/job/54143870206

### DIFF
--- a/.github/workflows/hardware-test.yml
+++ b/.github/workflows/hardware-test.yml
@@ -33,8 +33,9 @@ jobs:
       uses: actions/checkout@v4
 
     - name: Install build dependencies
+      shell: bash
       run: |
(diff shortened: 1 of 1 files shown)


alewycky-tenstorrent:
This is fine, but also there is the poorly-documented option `-o DPkg::Lock::Timeout=-1`, [implemented here.](https://salsa.debian.org/apt-team/apt/-/merge_requests/109/diffs?commit_id=93e1565796b61eb44bec39f50e09a34cbe090178)
(Also I wonder why it only waited for 12 seconds when the code says infinite or 120s.)

PR #165, opened 2025-11-01
ci: Simplify and fix Debian packaging system

joelsmithTT:
Replace the debhelper-based packaging to fix bugs and poor performance.

Changes:
- Fix version truncation bug: 2.5.0-rc1 now correctly creates /usr/src/tenstorrent-2.5.0-rc1/ instead of /usr/src/tenstorrent-2.5.0/
- Reduce packaging workflow time from 5+ minutes to ~20 seconds
- Build single universal .deb instead of 3 per-distro packages
- Add tools/build_debs.sh for local package builds
- Remove debian/ folder to reduce maintenance surface area
- Add version sanity check preventing module.h/dkms.conf mismatches
- Remove non-functional package signing implementation

Tested: Package builds successfully, installs correctly via dpkg, and DKMS registers with the full version string preserved.


tools/build_debs.sh
```diff
@@ -0,0 +1,133 @@
...
+    echo "  dkms.conf: ${DKMS_VERSION}" >&2
+    echo "" >&2
+    echo "These must match. Please update dkms.conf or module.h to match." >&2
+    exit 1
+fi
+
+echo "Building ${PACKAGE_NAME} version ${VERSION}"
+
+# Create temporary build directory
+BUILD_DIR=$(mktemp -d -t tenstorrent-deb-build.XXXXXX)
```
alewycky-tenstorrent: [mktemp manual says -t is deprecated.](https://www.gnu.org/software/coreutils/manual/coreutils.html#index-_002dt-19) Please use `--tmpdir` (with no value) instead.

PR #166, opened 2025-11-02
Fix NULL pointer dereference in telemetry init

joelsmithTT:
The sysfs attribute arrays were not NULL-terminated, causing the kernel's sysfs code to read past the end of the array during device initialization. This resulted in a kernel oops when loading the module on ARM64:

Unable to handle kernel NULL pointer dereference at virtual address 000000000000000a Call trace:
  create_files+0xc0/0x248 (P)
  internal_create_group+0x1f4/0x368
  sysfs_create_group+0x24/0x50
  devm_device_add_group+0x5c/0xf8
  blackhole_init_telemetry+0x1ec/0x2d0 [tenstorrent]
  tenstorrent_pci_probe+0x2b0/0x390 [tenstorrent]

The sysfs API requires attribute arrays to be NULL-terminated so it knows where to stop iterating. Fix by allocating space for one additional pointer in the telemetry_attrs arrays (the extra slot is zeroed by devm_kzalloc).

Affected both blackhole and wormhole device initialization.

### DIFF
--- a/blackhole.c
+++ b/blackhole.c
@@ -848,9 +848,9 @@ static bool blackhole_init(struct tenstorrent_device *tt_dev)
 	struct device *dev = &tt_dev->pdev->dev;
 	int i;
 
-	bh->hwmon_attr_addrs = devm_kzalloc(dev, sizeof(u64) * ARRAY_SIZE(bh_hwmon_attrs), GFP_KERNEL);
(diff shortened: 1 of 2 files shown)


alewycky-tenstorrent:
Convert all the telemetry/sysfs array allocations to devm_kcalloc.

Issue #174, opened 2025-11-22
[Feature]: KMD blocks BAR & TLB mapping access after reset


alewycky-tenstorrent:
### Is your feature request related to a problem? Please describe.

Applications can create memory mappings of BARs and TLBs and access them during/after board reset:
- In a VM scenario, if the host observes hotunplug, it may remove the mappings to the device from the guest's nested page tables. Access to the device triggers a fault to the hypervisor which shuts down the guest.
- TLB state is wiped during reset. Everybody ends up accessing NOC 0-0 @ 0.

### Describe the Solution You'd Like

Cause BAR & TLB mappings to fault during & after reset. Unmapping alone is not ideal because removing the virtual address allocation allows a concurrent mmap to reclaim those addresses. The best would be to somehow convert to PROT_NONE as if by mprotect.

### Describe Alternatives You've Considered

Migration to the IO ioctls may help but it won't cover all use-cases yet.

### Why is this Feature Important?

1. Mistakes happen. We want to reduce the blast radius of a programming mistake.
2. Long-running processes such as telemetry may not know that a concurrent reset is happening.

### Proposed Design/Technical Details (Optional)

Joel has a page zapping prototype.

### Use Cases

Applies to all user-mode software that maps BARs or TLBs, i.e. all SW that accesses the HW today. It is a robustness improvement.

joelsmithTT:
Zapping prototype, modeled after VFIO's technique: https://github.com/tenstorrent/tt-kmd/commit/1278d44b0e1c2a141d4105b34731226ed441e896

PR #183, opened 2025-11-25
Invalidate userspace mappings on device reset

joelsmithTT:
This is an attempt to solve https://github.com/tenstorrent/tt-kmd/issues/174; see the commit message for implementation details.

### DIFF
--- a/memory.c
+++ b/memory.c
@@ -12,6 +12,7 @@
 #include <linux/iommu.h>
 #include <linux/file.h>
 #include <linux/vmalloc.h>
+#include <linux/sched/mm.h>
 
 #include "chardev_private.h"
 #include "device.h"
@@ -952,6 +953,7 @@ long ioctl_allocate_tlb(struct chardev_private *priv,
 long ioctl_free_tlb(struct chardev_private *priv, struct tenstorrent_free_tlb __user *arg) {
 	struct tenstorrent_device *tt_dev = priv->device;
 	struct tenstorrent_free_tlb_in in = {0};
+	struct tenstorrent_mmap_vma *mmap_vma;
 	int ret;
 
 	if (copy_from_user(&in, &arg->in, sizeof(in)))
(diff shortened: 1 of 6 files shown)


memory.c
```diff
@@ -1317,3 +1338,77 @@ void tenstorrent_memory_cleanup(struct chardev_private *priv)
...
 }
+
+// Compatibility for kernels < 5.8 (mmap_sem vs mmap_lock)
+#if LINUX_VERSION_CODE < KERNEL_VERSION(5, 8, 0)
+#define mmap_read_lock(mm)       down_read(&(mm)->mmap_sem)
+#define mmap_read_unlock(mm)     up_read(&(mm)->mmap_sem)
+#endif
+// Zap all TLB and BAR mappings for this device, to prevent userspace from
+// accessing device memory after the device has been reset.
+void tenstorrent_vma_zap(struct tenstorrent_device *tt_dev)
```
alewycky-tenstorrent: This code groups zaps by mm, but I don't understand why that's important. I think it's much simpler to pop the head off priv->vma_list and zap vmas one-by-one.

memory.c
```diff
@@ -1039,34 +1047,75 @@ static struct dmabuf *vma_dmabuf_target(struct chardev_private *priv,
...
-		mutex_lock(&priv->mutex);
-		list_del(&mapping->list);
-		mutex_unlock(&priv->mutex);
+	priv = vma->vm_file->private_data;
+	if (!priv)
+		return;
 
-		kfree(mapping);
+	// Don't trust vm_private_data; it may be stale if tenstorrent_vma_zap()
+	// already freed the structure. Search to see if it still exists.
```
alewycky-tenstorrent: (I gather this happens because vm_ops->close may be called while only holding a mmap read lock.)

Instead, how about tenstorrent_vma_zap calls list_del_init and you defer kfree to tenstorrent_vma_close?

Issue #178, opened 2025-11-23
[Feature]: Implement blocking support for LOCK_CTL IOCTL


joelsmithTT:
### Is your feature request related to a problem? Please describe.

The current `LOCK_CTL` IOCTL only supports non-blocking lock acquisition. Implementing the C++ `Lockable` concept in userspace requires polling loops, which waste CPU cycles and introduce latency.

### Describe the Solution You'd Like

Augment the driver to support blocking lock acquisition. The kernel should put the calling process to sleep until the requested lock index is available, then wake it up.

### Describe Alternatives You've Considered

Userspace polling with `usleep` or `sched_yield`. This is inefficient and imprecise.

### Why is this Feature Important?

Essential for standard C++ synchronization primitives (`std::lock_guard`, `std::unique_lock`) without busy-waiting. Reduces CPU overhead for contended locks.

### Proposed Design/Technical Details (Optional)

1.  Add `TENSTORRENT_LOCK_CTL_ACQUIRE_BLOCKING` command to `ioctl.h`.
2.  Add `wait_queue_head_t resource_wait_queues[TENSTORRENT_RESOURCE_LOCK_COUNT]` to `struct tenstorrent_device` in `device.h`.
3.  Initialize wait queues in `tenstorrent_register_device`.
4.  In `ioctl_lock_ctl`, use `wait_event_interruptible` for blocking acquire.
5.  In `ioctl_lock_ctl`, call `wake_up_interruptible` on release.

### Use Cases

1.  **Simplification of Inter-Process Synchronization**: The UMD currently implements a shared mutex scheme using shared memory and `PTHREAD_MUTEX_ROBUST` to handle multi-process coordination. This kernel-level blocking mechanism would provide a simpler, centralized alternative for locking device resources that inherently handles process termination cleanup without complex userspace logic.
2.  **Cross-Language/Cross-Runtime Resource Arbitration**: Applications contending for hardware resources not managed by the KMD (e.g., specific ERISC cores) need a common synchronization point. A performant blocking IOCTL encourages adoption of a voluntary locking scheme across disparate clients (UMD, Luwen, direct IOCTL users) by eliminating the performance penalty of polling.

### Additional Context

A convention mapping hardware resources (e.g., ERISC core indices) to specific `LOCK_CTL` lock indices is required. The KMD should establish and define these mappings as constants to ensure all clients coordinate correctly.

PR #184, opened 2025-11-26
Implement blocking lock acquisition in LOCK_CTL

joelsmithTT:
Introduces blocking semantics to the LOCK_CTL ioctl, allowing userspace applications to wait efficiently for hardware locks without polling. Also cleans up the test suite and establishes conventions for lock usage.

Closes https://github.com/tenstorrent/tt-kmd/issues/178

### DIFF
--- a/test/lock.cpp
+++ b/test/lock.cpp
@@ -1,10 +1,18 @@
 // SPDX-FileCopyrightText: © 2024 Tenstorrent Inc.
 // SPDX-License-Identifier: GPL-2.0-only
 
+#include <atomic>
 #include <cerrno>
+#include <chrono>
 #include <cstdint>
+#include <mutex>
+#include <thread>
 
+#include <signal.h>
 #include <sys/ioctl.h>
+#include <sys/time.h>
+#include <sys/wait.h>
+#include <unistd.h>
 
 #include "ioctl.h"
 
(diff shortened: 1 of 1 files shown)


test/lock.cpp
```diff
@@ -1,174 +1,347 @@
...
-    zero(&lock);
-    lock.in.output_size_bytes = sizeof(lock.out);
+// Lock state bits returned by query_lock()
+static constexpr uint8_t LOCK_LOCAL  = 0b01;  // This fd holds the lock
+static constexpr uint8_t LOCK_GLOBAL = 0b10;  // Some fd holds the lock
 
-    lock.in.flags = TENSTORRENT_LOCK_CTL_ACQUIRE;
-    lock.in.index = index;
+// Returns true if lock was acquired, false if already held by someone.
+bool acquire_lock(DevFd &dev, uint8_t index)
```
broskoTT: So the interface is using file descriptor, and an arbitrary index?
I can imagine anyone using this to want to have a string based locks... But I guess it would then be up to user code to create a constant map between strings and ints...

test/lock.cpp
```diff
@@ -1,174 +1,347 @@
...
+    if (!release_lock(fd0, 0))
+        THROW_TEST_FAILURE("fd0 should release lock 0");
+
+    // 2. Can't release an unheld lock.
+    if (release_lock(fd0, 0))
+        THROW_TEST_FAILURE("fd0 released lock 0 but didn't hold it");
+
+    // 3. Can't release another fd's lock.
+    if (!acquire_lock(fd0, 0))
+        THROW_TEST_FAILURE("fd0 should acquire lock 0");
```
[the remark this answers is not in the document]
joelsmithTT: I aim for stylistic consistency with the rest of the test suite.  I'd prefer to maintain that consistency rather than introduce a new pattern here (or introduce a new pattern across all the tests).

device.h
```diff
@@ -44,6 +45,7 @@ struct tenstorrent_device {
 	struct notifier_block reboot_notifier;
 
 	DECLARE_BITMAP(resource_lock, TENSTORRENT_RESOURCE_LOCK_COUNT);
+	wait_queue_head_t resource_wait_queues[TENSTORRENT_RESOURCE_LOCK_COUNT];
```
alewycky-tenstorrent: Please use a single wait queue for all locks. (I think this comes out to 1.5kB for normal kernels and 3kB for PREEMPT_RT.)

test/lock.cpp
```diff
@@ -198,11 +216,204 @@ void VerifyAllLocks(const EnumeratedDevice &dev)
...
+    void lock() { acquire_lock_blocking(dev_, index_); }
+    bool try_lock() { return acquire_lock(dev_, index_); }
+
+    // unlock() shouldn't throw per BasicLockable, but for test code a clear
+    // failure beats silent misbehavior.
+    void unlock()
+    {
+        if (!release_lock(dev_, index_))
+            THROW_TEST_FAILURE("DeviceLock::unlock() failed");
+    }
```
alewycky-tenstorrent: Okay but I think you should check [std::uncaught_exception(s)](https://en.cppreference.com/w/cpp/error/uncaught_exception.html) before throwing so that you don't throw from inside an unwinding destructor.

joelsmithTT: I added it but it only helps if there's an exception already in flight.  It's still not legal to throw your way out of a destructor.  Maybe this should just call abort()?

PR #185, opened 2025-12-01
Add power policy for automatic low-power state management

joelsmithTT:
This change introduces a driver-side power policy layer on top of the existing SET_POWER_STATE ioctl, enabling automatic power management with safe defaults for backward compatibility.

Key features:

1. O_APPEND at open(): Signals that the client will explicitly manage power via TENSTORRENT_IOCTL_SET_POWER_STATE. These clients start with all power flags OFF (validity=0xF, flags=0) and no aggregation is triggered at open.

2. Legacy Clients: Clients that don't use O_APPEND default to high power (all flags ON except AI_CLK) and trigger immediate aggregation. This ensures backward compatibility, including for future firmware power domains (all 15 flag bits enabled by default).

3. power_policy module parameter (default=on):
   - At probe: Initialize device to low power via aggregation.
   - At close: Remove the closing fd's contribution and re-aggregate. When the last client closes, the device returns to low power.
   - Can be disabled at load (modprobe tenstorrent power_policy=0) or runtime (echo 0 > /sys/module/tenstorrent/parameters/power_policy) for compatibility with older firmware that doesn't safely handle power-down sequencing.

4. Documentation: Updates ioctl.h to document the new behavior.

5. Update power.c
  - Use O_APPEND to signal explicit power management.
  - Wait for user input before close (contribution removed on close).
  - Add TT_POWER_FLAG_TENSIX_ENABLE and TT_POWER_FLAG_L2CPU_ENABLE definitions.

### DIFF
--- a/chardev.c
+++ b/chardev.c
@@ -26,6 +26,8 @@
 #include "module.h"
 #include "tlb.h"
 
+#define TT_POWER_FLAG_ALL 0x7FFF
+
 static dev_t tt_device_id;
 static struct class *tt_dev_class;
 static unsigned int tt_max_devices;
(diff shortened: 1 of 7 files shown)


chardev.c
```diff
@@ -561,6 +563,7 @@ static int tt_cdev_open(struct inode *inode, struct file *file)
 {
 	struct tenstorrent_device *tt_dev = inode_to_tt_dev(inode);
 	struct chardev_private *private_data;
+	bool legacy_client = !(file->f_flags & O_APPEND);
```
alewycky-tenstorrent: Give this a name that says what capability is different, maybe power_aware?

PR #190, opened 2025-12-18
Use host PID in debugfs/procfs output

joelsmithTT:
Use task_tgid_nr() instead of task_tgid_vnr() to record the PID in the initial (host) namespace rather than the process's own namespace.


alewycky-tenstorrent:
What's the problem with current ns relative pid? Is it not always defined? Or do users really always want the init namespace pid?

Issue #193, opened 2026-01-09
[Bug]: Kernel oops in wormhole_save_reset_state when BAR regions are unassigned


joelsmithTT:
### Describe the Bug

Kernel oops during device probe when a Wormhole card has unassigned PCI BAR memory regions. The driver calls save_reset_state() unconditionally in tenstorrent_pci_probe(), even when init_device() fails because pci_iomap() returned NULL for the unassigned BARs. This causes a page fault when attempting to write to the NULL-based bar4_mapping.

### Steps to Reproduce

1. Have a Wormhole card with unassigned PCI BARs (e.g. connected via Thunderbolt)
2. Load the tenstorrent module
3. Kernel oops

### Expected Behavior

The driver should gracefully fail to probe devices with unassigned BAR regions without crashing the kernel.

### Actual Behavior

```
BUG: unable to handle page fault for address: 0000000001f30078
#PF: supervisor write access in kernel mode
RIP: 0010:iowrite32+0x9/0x80
Call Trace:
 wormhole_save_reset_state+0x40/0x130 [tenstorrent]
 tenstorrent_pci_probe+0x2bc/0x3f0 [tenstorrent]
```

### Kernel Information

6.17.0

### Driver Information

2.6.0

### Hardware Information

Specific Device Affected: n150 (Wormhole) connected via Thunderbolt
CPU: Intel Core Ultra 9 285K

### Relevant Logs and Output

```
    8a:00.0 Processing accelerators: Tenstorrent Inc Wormhole (rev 01)
            Subsystem: Tenstorrent Inc n150
            Memory at <unassigned> (64-bit, prefetchable)
            Memory at b3f00000 (32-bit, non-prefetchable) [size=1M]
            Memory at <unassigned> (64-bit, prefetchable)
```

PR #194, opened 2026-01-09
Fail gracefully when PCI BARs are unassigned

joelsmithTT:
When a device has unassigned BAR regions the driver would previously
crash:

- Probe called save_reset_state() unconditionally, causing a page fault
  when bar4_mapping was NULL (Wormhole)
- TLB mmap would attempt to map physical address 0

Now probe fails with -EIO if init_device() fails, and TLB mmap returns
-ENXIO for unassigned or unexpectedly shrunken BARs.

Fixes https://github.com/tenstorrent/tt-kmd/issues/193

### DIFF
--- a/memory.c
+++ b/memory.c
@@ -1215,6 +1215,11 @@ static int map_tlb_window(struct chardev_private *priv, struct vm_area_struct *v
 		goto unlock;
 	}
 
+	if (tlb_desc.bar_offset + size > pci_resource_len(tt_dev->pdev, tlb_desc.bar)) {
+		ret = -ENXIO;
+		goto unlock;
+	}
+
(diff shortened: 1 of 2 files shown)


memory.c
```diff
@@ -1215,6 +1218,11 @@ static int map_tlb_window(struct chardev_private *priv, struct vm_area_struct *v
 		goto unlock;
 	}
 
+	if (pci_resource_len(tt_dev->pdev, tlb_desc.bar) == 0) {
+		ret = -ENXIO;
+		goto unlock;
+	}
+
```
alewycky-tenstorrent: Please check if `tlb_desc.bar_offset + size` fits. That way we're protected against unexpectedly shrunken BARs. (Or dynamically compute tlb_counts for all classes? But it's a hassle on WH.)

PR #195, opened 2026-01-09
Fix user-pinned DMA for Thunderbolt-connected devices

joelsmithTT:
The kernel marks Thunderbolt-connected devices as "untrusted", forcing DMA through SWIOTLB bounce buffering. This breaks user-pinned DMA: the SWIOTLB path allocates IOVAs per scatter-gather entry rather than as a contiguous range, so PIN_PAGES fails the contiguity check. Even if it didn't fail there, bounce buffering is fundamentally incompatible with user-pinned memory since the device would DMA to bounce buffers rather than the user's pages.

Clear the untrusted flag during probe. With IOMMU active, the device is already restricted to explicitly-mapped addresses and cannot access arbitrary system memory.

Also add fix-tt-hotplug-bars, a helper script to work around BAR allocation failures on Thunderbolt hotplug. Requires "pci=realloc" on the kernel command line.

### DIFF
--- a/enumerate.c
+++ b/enumerate.c
@@ -278,6 +278,21 @@ static int tenstorrent_pci_probe(struct pci_dev *dev, const struct pci_device_id
 	dma_set_max_seg_size(&dev->dev, UINT_MAX);
 	dma_set_seg_boundary(&dev->dev, ULONG_MAX);
 
+	// Thunderbolt-connected devices are marked "untrusted" by the kernel,
+	// which forces SWIOTLB bounce buffering even with IOMMU enabled. The
+	// IOMMU itself provides DMA isolation; clear the flag only when IOMMU
+	// is active and the untrust is due to an external-facing port.
+	// pci_dev.external_facing was added in v5.9.
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(5, 9, 0)
+	if (dev->untrusted && is_iommu_translated(&dev->dev)) {
+		struct pci_dev *root = pcie_find_root_port(dev);
+		if (root && root->external_facing) {
+			dev_info(&dev->dev, "Clearing untrusted flag (external-facing port, IOMMU active)\n");
+			dev->untrusted = 0;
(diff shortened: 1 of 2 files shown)


enumerate.c
```diff
@@ -264,6 +264,14 @@ static int tenstorrent_pci_probe(struct pci_dev *dev, const struct pci_device_id
 	dma_set_max_seg_size(&dev->dev, UINT_MAX);
 	dma_set_seg_boundary(&dev->dev, ULONG_MAX);
 
+	// Thunderbolt-connected devices are marked "untrusted" by the kernel, which
+	// forces SWIOTLB bounce buffering even with IOMMU enabled. Clear the flag
+	// since we trust our own hardware.
+	if (dev->untrusted) {
+		dev_info(&dev->dev, "Clearing untrusted flag\n");
+		dev->untrusted = 0;
```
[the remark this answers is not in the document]
alewycky-tenstorrent: Also we only want to trust if the untrust is due to being an external device. The kernel doesn't actually say why but maybe we shouldn't trust if it's not external since then we don't know why it's untrusted.

joelsmithTT:
VFIO does not appear to have this issue because it bypasses the DMA API.

tools/fix-tt-hotplug-bars
```diff
@@ -0,0 +1,106 @@
...
+        if [[ $DRY_RUN -eq 1 ]]; then
+            log "[dry-run] would remove $bridge and rescan"
+        else
+            echo 1 > "/sys/bus/pci/devices/$bridge/remove"
+            sleep 2
+            echo 1 > /sys/bus/pci/rescan
+            sleep 3
+        fi
+        
+        ((fixed++)) || true
```
alewycky-tenstorrent: Maybe check has_unassigned_bars again to make sure it really got fixed?

joelsmithTT: Added above

PR #234, opened 2026-04-06
Use a sparse representation for the telemetry tag cache.

alexapostolu:
Resolves #233,
> Use a dynamically-allocated sorted array of (tag_id, address) pairs instead of a flat tag-ID-indexed array.
> Removes the fixed upper bound on tag IDs and savess ~1 KiB from the per-device struct.


alexapostolu:
Oh I see some tests are failing to compile, what C version should I be compiling with locally?

telemetry.c
```diff
@@ -1,18 +1,63 @@
...
+			return true;
+	}
+
+	for (i = 0; tt_dev->telemetry_sysfs[i].tag_id != 0; i++) {
+		const struct tenstorrent_sysfs_attr *attr = &tt_dev->telemetry_sysfs[i];
+
+		if (attr->tag_id == tag_id)
+			return true;
+		if (attr->attr.show == tt_sysfs_show_u64_hex && attr->tag_id + 1 == tag_id)
+			return true;
```
joelsmithTT, line 54: I forgot about the hidden dependency of BOARD_ID and ASIC_ID on their other halves.  This probably deserves a comment.

wormhole.c
```diff
@@ -380,6 +382,7 @@ static struct tenstorrent_sysfs_attr wh_sysfs_attributes[] = {
 	{ TELEMETRY_ETH_FW_VERSION, __ATTR(tt_eth_fw_ver, S_IRUGO, tt_sysfs_show_u32_ver, NULL) },
 	{ TELEMETRY_ASIC_ID, __ATTR(tt_asic_id, S_IRUGO, tt_sysfs_show_u64_hex, NULL) },
 	{ TELEMETRY_TIMER_HEARTBEAT, __ATTR(tt_heartbeat, S_IRUGO, tt_sysfs_show_u32_dec, NULL) },
+	{ 0 },
```
joelsmithTT: In `wormhole_init` the `telemetry_attrs` array is allocated based on the size of this array, plus one.  Same for blackhole.c

wormhole.c
```diff
@@ -762,6 +785,12 @@ static void wormhole_cleanup_telemetry(struct tenstorrent_device *tt_dev)
 {
 	struct wormhole_device *wh_dev = tt_dev_to_wh_dev(tt_dev);
 
+	cancel_delayed_work_sync(&wh_dev->fw_ready_work);
```
joelsmithTT, line 795: `tenstorrent_pci_remove` has a hack that does this, but it happens earlier (prior to `cleanup_hardware`).
I wonder if `tenstorrent_pci_remove` should be reordered so `cleanup_telemetry` runs before `cleanup_hardware`.

blackhole.c
```diff
@@ -483,17 +485,36 @@ static int telemetry_probe(struct tenstorrent_device *tt_dev)
 	}
 
 	num_entries = noc_read32(bh, ARC_X, ARC_Y, base_addr + 4, 0);
+	if (num_entries == 0) {
+		dev_warn(&tt_dev->pdev->dev, "No telemetry entries found.\n");
+		return 0;
+	}
+
+	cache = kcalloc(num_entries, sizeof(*cache), GFP_KERNEL);
```
joelsmithTT, line 493: This ends up over-allocating.  Number of telemetry items published by FW vs number of items we care about.

telemetry.c
```diff
@@ -1,18 +1,67 @@
...
+			sizeof(*tt_dev->telemetry_cache), telem_cache_entry_cmp);
+
+	return found ? found->address : 0;
+}
+
+bool telem_tag_is_wanted(const struct tenstorrent_device *tt_dev, u16 tag_id)
+{
+	u16 i;
+
+	if (!tt_dev->hwmon_attributes || !tt_dev->telemetry_sysfs_count)
```
joelsmithTT: Not a problem because both architectures have both types of attribute, but this makes the function return false fir all tags if either table is absent.

blackhole.c
```diff
@@ -483,17 +485,36 @@ static int telemetry_probe(struct tenstorrent_device *tt_dev)
...
+	if (!cache) {
+		dev_err(&tt_dev->pdev->dev, "Telemetry cache could not be allocated.\n");
```
joelsmithTT: I suggest shortening this to just the return (we don't log allocation failure anywhere else).

blackhole.c
joelsmithTT: ditto for wormhole.c

blackhole.c
joelsmithTT: ditto for wormhole.c

blackhole.c
alexapostolu: Is realloc fine to use here?

joelsmithTT: yes

wormhole.c
```diff
@@ -534,13 +532,17 @@ static int wormhole_read_telemetry_tag(struct tenstorrent_device *tt_dev, u16 ta
...
+	struct telem_cache_entry *cache;
 	u32 base_addr, data_addr;
 	u32 version, major_ver, minor_ver, patch_ver;
 	u32 tags_addr;
 	u32 num_entries;
 	u32 i;
+	u16 count = 0;
 
-	memset(tt_dev->telemetry_tag_cache, 0, sizeof(tt_dev->telemetry_tag_cache));
+	kfree(tt_dev->telemetry_cache);
```
joelsmithTT, line 543: There's a use-after-free here: POST_RESET vs telemetry read.  The workqueue thread for WH's slow telemetry init is implicated:


Step | POST_RESET | fw_ready_work_func | sysfs read
-- | -- | -- | --
1 | holds reset_rwsem (write); wormhole_probe_telemetry, schedule_delayed_work on fw_ready_work | queued | blocked on reset_rwsem
2 | returns, drops reset_rwsem | queued | acquires reset_rwsem (read), enters bsearch on telemetry_cache
3 | — | runs, calls telemetry_probe, kfree on telemetry_cache | still bsearch-ing telemetry_cache
4 | — | kcalloc may hand memory back to another subsystem | dereferences telemetry_cache: use-after-free

I just checked on a newish FW and although the telemetry init time is not as slow as it used to be, it is still slow enough that we probably shouldn't do it synchronously (although I am open to changing my mind about this -- @alewycky-tenstorrent any thoughts?)  It's still north of 2 seconds.

Does this fix it?
```
diff --git a/wormhole.c b/wormhole.c
index 4d85164..d6c0ed6 100644
--- a/wormhole.c
+++ b/wormhole.c
@@ -691,7 +691,9 @@ static void fw_ready_work_func(struct work_struct *work)
                return;
        }
 
+       down_write(&tt_dev->reset_rwsem);
        telemetry_probe(tt_dev);
+       up_write(&tt_dev->reset_rwsem)
 
        // First-time init: register sysfs and hw
```

alewycky-tenstorrent:
Take this change even further: Create HW-independent telemetry_probe that
1. Creates a telemetry cache with only tags but no addresses. (Sorted by tag_id.)
2. Passes that to a new tt_dev->populate_telemetry_cache() that fills in the addresses.
3. Scans the addresses for any that are invalid (Need a magic number say 0xF...) and squeezes those entries from the list
4. Replaces the existing telemetry cache with the new one.

Sorting in step 1 solves two problems: it allows binary search later and it simplifies populate_telemetry_cache. (If  you implement PTC by reading each tag from CSM, then you only need to check the next cache entry to see if that tag is wanted.)

alewycky-tenstorrent:
Lift the call to telem_cache_lookup from blackhole/wormhole_read_telemetry_tag to tt_telemetry_read32. Then tt_telemetry_read32 passes an "address" down to HW-specific functions instead of a tag ID.

alewycky-tenstorrent:
Create tt_telemetry_cleanup in telemetry.c to reverse the work of tt_telemetry_populate. It will call the HW-specific function and kfree the telemetry cache (lift that from HW-specific into HW-independent). tenstorrent_pci_remove will call tt_telemetry_cleanup.

alewycky-tenstorrent (approved):

PR #236, opened 2026-05-04
Enable Wormhole power management with deferred idle powerdown

joelsmithTT:
## Summary

GDDR is the dominant component of WH idle power; gating it saves~630 W per Galaxy (SYS-2973).  This series turns on the KMD plumbing that lets firmware drive that gating from fd ownership.

- Re-enable the FW path in `wormhole_set_power_state` (was `#if 0` pending FW work) and bump `ARC_MSG_TIMEOUT_MS` to 1000.
- Defer the synchronous idle-powerdown from last-close to a delayed work item with a configurable grace period.  An `open()` arriving inside the window cancels the pending powerdown, so a fast-following open doesn't pay the GDDR-on cost (17 ms - 800 ms per chip on current FW, dominated by stress-trained channels).

Default-off for the deferral: `idle_power_down_grace_ms=0` falls back to synchronous-at-close.  Operators opt in via the module param or `/sys/module/tenstorrent/parameters/idle_power_down_grace_ms`.

## Firmware
The handler in FW 19.8.x is unsafe to drive (PCIe speed-change side effects; Tensix/GDDR ordering); subsequent FW fixes both.  This PR runs against any WH firmware with the corrected handler.  In-flight performance mitigations (skip-slow channel; Samsung GDDR exclusion) improve the experience but aren't required for correctness.

~Forward-compat with old FW is handled by a separate V2-opcode change in both FW and KMD landing before customer release; out of scope here.~

### DIFF
--- a/chardev.c
+++ b/chardev.c
@@ -441,6 +441,13 @@ int tenstorrent_set_aggregated_power_state(struct tenstorrent_device *tt_dev)
 	return ret;
 }
 
+// Delayed work: send the aggregated idle power-down message after the
+// grace period elapsed with no open fds.  Body added in a later change;
+// this stub is present so teardown can cancel the work unconditionally.
+void tenstorrent_power_down_work_func(struct work_struct *work)
+{
+}
+
 static long ioctl_set_power_state(struct chardev_private *priv, struct tenstorrent_power_state __user *arg)
 {
 	struct tenstorrent_device *tt_dev = priv->device;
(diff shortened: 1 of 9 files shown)


joelsmithTT:
Credit to @broskoTT for the deferred powerdown idea.

If we ship this, we need to decide whether to pick a reasonable default idle_power_down_grace_ms (five minutes?) or leave it at zero and have it be opt-in.

njelicTT (approved):

PR #237, opened 2026-05-05
Fix LOCK_CTL ACQUIRE_BLOCKING vs RESET_DEVICE deadlock

### LINKED ISSUE #235: LOCK_CTL ACQUIRE_BLOCKING vs RESET_DEVICE deadlock on reset_rwsem
### Describe the Bug

TENSTORRENT_IOCTL_LOCK_CTL with ACQUIRE_BLOCKING sleeps in wait_event_interruptible() while holding reset_rwsem for read.  A subsequent TENSTORRENT_IOCTL_RESET_DEVICE blocks indefinitely in down_write(reset_rwsem).

### Steps to Reproduce

https://gist.github.com/joelsmithTT/74d790561823f617bc7dacaf6191aaef

joelsmithTT:
tt_cdev_ioctl() takes reset_rwsem for read on every non-reset ioctl and for write on RESET_DEVICE. ioctl_lock_ctl()'s ACQUIRE_BLOCKING path parked the caller in wait_event_interruptible() while still holding down_read(reset_rwsem), so a subsequent RESET_DEVICE blocked indefinitely in down_write(). The rwsem is writer-fair, so a pending writer also blocked any concurrent LOCK_CTL_RELEASE that would otherwise have woken the parked waiter, making it a hard deadlock until the waiter took a signal. The same shape would deadlock against tenstorrent_pci_remove, which also takes the rwsem for write.

Extract acquire_resource_lock_blocking(), which drops reset_rwsem for the duration of the wait and reacquires it before returning so the caller's matching up_read stays balanced. Broaden the wait condition to also unblock when the device is detached or reset_gen has advanced past the fd's open_reset_gen, and wake the waitqueue from ioctl_reset_device and tenstorrent_pci_remove so parked waiters exit promptly with -ENODEV.

Resource locks are intentionally preserved across reset: a reset invalidates every pre-reset fd (open_reset_gen mismatch), so a holder cannot release via ioctl after a reset and only close(fd) clears the bits. The one carve-out is the case where a waiter wins test_and_set_bit() and a reset/remove then races in before reacquire: give the bit back, since the caller never observed a successful acquire on a fd that is now invalid.

Fixes: #235
Assisted-by: Claude:claude-4.7-opus

### DIFF
--- a/chardev.c
+++ b/chardev.c
@@ -264,6 +264,12 @@ static long ioctl_reset_device(struct chardev_private *priv,
 	out.output_size_bytes = sizeof(out);
 	out.result = !ok;
 
+	// Wake any LOCK_CTL ACQUIRE_BLOCKING waiters. If reset_gen was bumped
+	// or the device is otherwise unusable for them, they will observe it
+	// and return -ENODEV instead of waiting forever for a lock that no
+	// pre-reset fd can ever release via ioctl.
+	wake_up_interruptible(&tt_dev->resource_lock_waitqueue);
+
 	if (clear_user(&arg->out, in.output_size_bytes) != 0)
 		return -EFAULT;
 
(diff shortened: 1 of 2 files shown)


njelicTT (approved):
In general, the patch seems to make sense. Do we have any tests that validate the behaviour of the locks in this driver?

joelsmithTT:
> Do we have any tests that validate the behaviour of the locks in this driver?

Yes - see test/lock.cpp.
Also, https://github.com/tenstorrent/tt-kmd/issues/235 has a link to a standalone repro for the defect this PR fixes.

PR #241, opened 2026-05-18
Add O_EXCL exclusive-open support to the chardev

joelsmithTT:
## Motivation
tt-flash wants to flash and reset the device without any other process being able to use it, and without new processes being able to open it mid-flash.

See also: https://tenstorrent.atlassian.net/browse/PT-377

## Semantics
open() acts as a reader/writer lock: `O_EXCL` is the writer, plain opens are readers, and `O_NONBLOCK` selects trylock behavior.

| open flags | behavior |
|---|---|
| `O_RDWR` | Blocks (interruptible) while an `O_EXCL` fd is held, then proceeds once it is released; otherwise unchanged. |
| `O_RDWR \| O_NONBLOCK` | `-EAGAIN` while an `O_EXCL` fd is held; otherwise unchanged. |
| `O_RDWR \| O_EXCL` | Blocks (interruptible) until the device is idle, then takes exclusive ownership. |
| `O_RDWR \| O_EXCL \| O_NONBLOCK` | `-EAGAIN` if not immediately acquirable. |

While an `O_EXCL` fd is open, every other open blocks until it is released (or gets `-EAGAIN` with `O_NONBLOCK`). Applications that don't pass `O_EXCL` are unaffected unless something like tt-flash is actively holding the device, in which case their `open()` blocks until it finishes.

Fairness: this is a reader-preferring lock with barge-in (no FIFO ordering). Plain opens (readers) never starve. A blocking `O_EXCL` writer can be starved by a steady stream of plain opens, since it only wakes on the last-close transition; callers that can't tolerate that should use `O_EXCL | O_NONBLOCK` and retry. In tt-flash's case the system is quiesced before flashing, so there is no reader churn to starve against.

## Commits
1. Remove unused chardev open count and lifecycle callbacks — drops chardev_open_count, first_open_cb, last_release_cb, and the increment/decrement helpers that were never observed by anyone. open_fds_list already tracks who has the device open.

2. Add O_EXCL exclusive-open semantics for the chardev — implements the table above using a new chardev_excl_held bool and a waitqueue on tenstorrent_device, all gated by the existing chardev_mutex. Readers wait non-exclusively and writers wait exclusively on that waitqueue, so a single wake on the last-close transition releases every blocked reader plus one writer. Also folds in a reset-path fix: ioctls that bump reset_gen (CONFIG_WRITE, USER_RESET, ASIC_RESET, ASIC_DMC_RESET) now carry the issuing fd's open_reset_gen along with the bump, so the resetter's own fd stays valid for the follow-up POST_RESET. Other fds still mismatch and continue to fail with -ENODEV. Without this, tt-flash would have to close and reopen between ASIC_RESET and POST_RESET, dropping exclusivity in a window where another opener could interpose.

3. Add tests for O_EXCL chardev semantics — test/excl.cpp exercises plain coexistence, idle-acquire, EAGAIN for non-blocking opens (both plain-vs-O_EXCL and O_EXCL-vs-O_EXCL) while held, the release-restores-access transition, the blocking variants in both directions (an O_EXCL open waiting on a plain holder, and a plain open waiting on an O_EXCL holder — covering both an explicit close in another thread and implicit fd cleanup when the holder process exits), and EINTR on signal without SA_RESTART. Threading and signal patterns mirror the equivalents in lock.cpp.

### DIFF
--- a/chardev.c
+++ b/chardev.c
@@ -93,6 +93,7 @@ int tenstorrent_register_device(struct tenstorrent_device *tt_dev)
 	char name[16];
 
 	init_waitqueue_head(&tt_dev->resource_lock_waitqueue);
+	init_waitqueue_head(&tt_dev->chardev_excl_waitqueue);
 
 	device_initialize(&tt_dev->dev);
 	tt_dev->dev.devt = devt;
@@ -188,6 +189,14 @@ static long ioctl_get_driver_info(struct chardev_private *priv,
 	return 0;
 }
 
+// Bump the device's reset generation and bring this fd's open_reset_gen along
+// with it. Other fds become permanently invalid, but the resetter keeps a live
+// fd so it can complete the reset sequence without a close/reopen window.
+static void bump_reset_gen(struct chardev_private *priv)
+{
+	priv->open_reset_gen = atomic_long_inc_return(&priv->device->reset_gen);
+}
+
 static long ioctl_reset_device(struct chardev_private *priv,
 			       struct tenstorrent_reset_device __user *arg)
 {
@@ -220,21 +229,21 @@ static long ioctl_reset_device(struct chardev_private *priv,
 		tenstorrent_vma_zap(tt_dev);
 		ok = pcie_hot_reset_and_restore_state(pdev);
 	} else if (in.flags == TENSTORRENT_RESET_DEVICE_CONFIG_WRITE) {
-		atomic_long_inc(&priv->device->reset_gen);
+		bump_reset_gen(priv);
 		tenstorrent_vma_zap(tt_dev);
 		ok = pcie_timer_interrupt(pdev);
 	} else if (in.flags == TENSTORRENT_RESET_DEVICE_USER_RESET) {
-		atomic_long_inc(&priv->device->reset_gen);
+		bump_reset_gen(priv);
 		tenstorrent_vma_zap(tt_dev);
 		ok = set_reset_marker(pdev);
 		priv->device->needs_hw_init = true;
 	} else if (in.flags == TENSTORRENT_RESET_DEVICE_ASIC_RESET) {
-		atomic_long_inc(&priv->device->reset_gen);
+		bump_reset_gen(priv);
 		tenstorrent_vma_zap(tt_dev);
 		ok = priv->device->dev_class->reset(priv->device, in.flags);
 		priv->device->needs_hw_init = true;
 	} else if (in.flags == TENSTORRENT_RESET_DEVICE_ASIC_DMC_RESET) {
-		atomic_long_inc(&priv->device->reset_gen);
+		bump_reset_gen(priv);
 		tenstorrent_vma_zap(tt_dev);
 		ok = priv->device->dev_class->reset(priv->device, in.flags);
 		priv->device->needs_hw_init = true;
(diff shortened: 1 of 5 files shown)


kaplanz:
My 2 cents here are this is semantically equivalent to a [read-write lock](https://en.wikipedia.org/wiki/Readers–writer_lock), which makes sense as long as only programs that actually need exclusive access use `O_EXCL`.

I think the **only** justifiable processes that should use `O_EXCL` should be programs that _require_ a reset for their core functionality, e.g.:

- `tt-flash`: flash firmware (must reset to load new firmware)
- `tt-mod`: modifying SPI parameters (must reset to load new parameters)
- `tt-smi -r` or other reset-only mechanisms, (but **not** regular `tt-smi`)

Long running programs should **never** hold `O_EXCL`, and if they need to perform a reset they should close their existing `O_RDWR`-only `fd` then try again as `O_RDWR | O_EXCL` and block.

alewycky-tenstorrent:
Ultimately how will this integrate with tt-telemetry? I just want to know whether the whole problem is being solved.

chardev.c
```diff
@@ -615,11 +624,46 @@ static struct tenstorrent_device *inode_to_tt_dev(struct inode *inode)
...
+
+	mutex_lock(&tt_dev->chardev_mutex);
+
+	if (want_excl) {
+		while (!list_empty(&tt_dev->open_fds_list)) {
+			if (nonblock) {
+				mutex_unlock(&tt_dev->chardev_mutex);
+				return -EAGAIN;
+			}
+			mutex_unlock(&tt_dev->chardev_mutex);
```
alewycky-tenstorrent, line 737: Move `mutex_unlock` ahead of `if(nonblock)` because it's the first thing done in both branches.

chardev.c
```diff
@@ -615,11 +624,46 @@ static struct tenstorrent_device *inode_to_tt_dev(struct inode *inode)
...
+			ret = wait_event_interruptible_exclusive(tt_dev->chardev_excl_waitqueue,
+								 list_empty(&tt_dev->open_fds_list));
```
alewycky-tenstorrent, line 741: Is it safe to run list_empty without holding the lock that controls modification of the list?

joelsmithTT: I think so, either it sees it empty or it doesn't at the instant it is looking.

joelsmithTT:
Updated based on review: plain `open()` now blocks (or `-EAGAIN` with `O_NONBLOCK`) while an `O_EXCL` fd is held, instead of failing fast with `-EBUSY`. So `O_NONBLOCK` — not `O_EXCL` — now decides block-vs-fail for a contending opener, making this a proper reader/writer lock at open time. Pushed the code, tests, and updated the PR description to match. `-EBUSY` is gone entirely.

joelsmithTT:
> Ultimately how will this integrate with tt-telemetry? I just want to know whether the whole problem is being solved.

tt-telemetry should not open with `O_EXCL`.

The integration story is on the tt-flash side.  When it goes to flash, it requests `O_EXCL`. If any process (a tt-telemetry instance included) currently has the device open, tt-flash can either quit with an error or warn and wait.  The KMD exposes /proc/driver/tenstorrent/N/pids so tt-flash can print something actionable.

> I think the only justifiable processes that should use O_EXCL should be programs that require a reset for their core functionality

Agreed on the policy.  One caveat on `tt-smi -r`: requiring `O_EXCL` there would break the current user expectation that a reset always succeeds.  I'm in favor of nudging users away from expecting that resetting a chip out from under a running process should work.  However I'd keep `tt-smi -r` as-is for now and treat the O_EXCL-for-reset adoption as a separate, deliberate decision per tool.

chardev.c
```diff
@@ -708,11 +717,60 @@ static struct tenstorrent_device *inode_to_tt_dev(struct inode *inode)
...
+			if (ret)
+				return ret;
+			mutex_lock(&tt_dev->chardev_mutex);
+		}
+		tt_dev->chardev_excl_held = true;
+	} else {
+		// Reader: wait until no O_EXCL fd is held.  Registered non-exclusively
+		// so all readers wake together when the holder releases.
+		while (tt_dev->chardev_excl_held) {
```
alewycky-tenstorrent, line 750: Note that this line does *not* need READ_ONCE because it's under the lock and there is no possible concurrent writer.

chardev.c
```diff
@@ -708,11 +717,60 @@ static struct tenstorrent_device *inode_to_tt_dev(struct inode *inode)
...
+			mutex_unlock(&tt_dev->chardev_mutex);
+			if (nonblock)
+				return -EAGAIN;
+			ret = wait_event_interruptible_exclusive(tt_dev->chardev_excl_waitqueue,
+								 list_empty(&tt_dev->open_fds_list));
+			if (ret)
+				return ret;
+			mutex_lock(&tt_dev->chardev_mutex);
+		}
+		tt_dev->chardev_excl_held = true;
```
alewycky-tenstorrent: This must be WRITE_ONCE because it may be concurrent with line 755. (i.e. the test in wait_event_interruptible).

chardev.c
```diff
@@ -708,11 +717,60 @@ static struct tenstorrent_device *inode_to_tt_dev(struct inode *inode)
...
+	} else {
+		// Reader: wait until no O_EXCL fd is held.  Registered non-exclusively
+		// so all readers wake together when the holder releases.
+		while (tt_dev->chardev_excl_held) {
+			mutex_unlock(&tt_dev->chardev_mutex);
+			if (nonblock)
+				return -EAGAIN;
+			ret = wait_event_interruptible(tt_dev->chardev_excl_waitqueue,
+						       !tt_dev->chardev_excl_held);
```
alewycky-tenstorrent: Access to `tt_dev_chardev_excl_held` must be READ_ONCE because it's unlocked and possibly concurrent with line 746.

chardev.c
```diff
@@ -853,6 +917,11 @@ static int tt_cdev_release(struct inode *inode, struct file *file)
 	// Must follow the list_del: release_power reads open_fds_list.
 	tt_cdev_release_power(priv);
 
+	if (list_empty(&tt_dev->open_fds_list)) {
+		tt_dev->chardev_excl_held = false;
```
alewycky-tenstorrent: This must be WRITE_ONCE because chardev_mutex is not held at line 755.

PR #244, opened 2026-05-21
Fix DKMS cleanup during Debian upgrades

jasondavies:
## Summary

Fix Debian package maintainer scripts so stale Tenstorrent DKMS registrations do not survive package upgrades and break future kernel updates.

The generated `prerm` now removes the package's DKMS registration on both `remove` and `upgrade`. The generated `postinst` also cleans up stale Tenstorrent DKMS entries whose `source/dkms.conf` is already missing, repairing systems affected by older packages.

## Motivation

A system upgraded from `tenstorrent-dkms` 2.7.0 to 2.8.0 was left with stale DKMS state:

```
/var/lib/dkms/tenstorrent/2.7.0/source -> /usr/src/tenstorrent-2.7.0
```

but /usr/src/tenstorrent-2.7.0 no longer existed.

On a later kernel update, Ubuntu's DKMS hooks failed with:

Error! Could not locate dkms.conf file.
/var/lib/dkms/tenstorrent/2.7.0/source/dkms.conf does not exist.

This prevented DKMS from rebuilding the current tenstorrent/2.8.0 module for the new kernel.

## Testing

- Ran bash -n tools/build_debs.sh

### DIFF
--- a/tools/build_debs.sh
+++ b/tools/build_debs.sh
@@ -79,6 +79,21 @@ cat > "${DEBIAN_DIR}/postinst" << 'EOF'
 #!/bin/sh
 set -e
 
+# Remove stale DKMS registrations left behind by older packages.  Once the
+# matching /usr/src tree is gone, dkms remove cannot recover the entry.
+for dkms_dir in /var/lib/dkms/tenstorrent/*; do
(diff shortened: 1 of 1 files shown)


joelsmithTT:
Thanks for the patch!

Issue #249, opened 2026-05-28
O_APPEND can change between open and close


alewycky-tenstorrent:
`fcntl(fd, F_SETFL, flags)` can change the `O_APPEND` flag on an already-open file descriptor. Mostly by luck there's no actual bug, but we should improve the code. `tt_cdev_release` checks `O_APPEND` and that contributes to `has_power_contribution`, but all that really matters is `is_minimum_power`.

Also since you can call `SET_POWER_STATE` on a non-power-aware fd, `O_APPEND` tells you nothing about the fd's power state.

PR #250, opened 2026-05-29
Untangle tt_cdev_release power gating

joelsmithTT:
Refactor the chardev release code path to improve clarity.

Split the release function into helpers and add a _priv variant of the aggregation to avoid dropping and re-taking the chardev_mutex.

Context: https://github.com/tenstorrent/tt-kmd/issues/249

### DIFF
--- a/chardev.c
+++ b/chardev.c
@@ -447,14 +447,13 @@ static long ioctl_set_noc_cleanup(struct chardev_private *priv,
 	return 0;
 }
 
-int tenstorrent_set_aggregated_power_state(struct tenstorrent_device *tt_dev)
+static int tenstorrent_set_aggregated_power_state_locked(struct tenstorrent_device *tt_dev)
 {
 	struct tenstorrent_power_state power_state = { 0 };
 	struct chardev_private *priv;
 	u8 max_settings_count = 0;
-	int ret;
 
-	mutex_lock(&tt_dev->chardev_mutex);
+	lockdep_assert_held(&tt_dev->chardev_mutex);
 
 	list_for_each_entry(priv, &tt_dev->open_fds_list, open_fd) {
 		u8 flags_count;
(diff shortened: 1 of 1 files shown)


chardev.c
```diff
@@ -786,86 +793,92 @@ static int tt_cdev_open(struct inode *inode, struct file *file)
...
+	decrement_cdev_open_count(tt_dev);
+	tenstorrent_memory_cleanup(priv);
+	tt_cdev_release_resource_locks(priv);
+	tt_cdev_release_tlbs(priv);
 
 	mutex_lock(&tt_dev->chardev_mutex);
+
 	list_del(&priv->open_fd);
-	last_close = list_empty(&tt_dev->open_fds_list);
+	tt_cdev_release_power(priv);
```
alewycky-tenstorrent, line 873: It's not a blocker, but the hidden dependency between these two lines is troubling. Before it was pretty clear - we did `list_del` and `list_empty` back-to-back.

chardev.c
```diff
@@ -447,14 +447,13 @@ static long ioctl_set_noc_cleanup(struct chardev_private *priv,
 	return 0;
 }
 
-int tenstorrent_set_aggregated_power_state(struct tenstorrent_device *tt_dev)
+static int tenstorrent_set_aggregated_power_state_priv(struct tenstorrent_device *tt_dev)
```
alewycky-tenstorrent: How about `tenstorrent_set_aggregated_power_state_locked`?

chardev.c
joelsmithTT: When I was writing the patch I debated how to deal with this.  I didn't want to put the `list_del` into `tt_cdev_release_power`.  I had an interim state where `last_close` was passed into `tt_cdev_release_power` but I rejected that on aesthetic grounds (but also because `tenstorrent_set_aggregated_power_state` walks the list).

I added a comment.

chardev.c
joelsmithTT: Sure; done.

alewycky-tenstorrent (approved):

Issue #248, opened 2026-05-28
Enable NIC/Blackhole peer-to-peer RDMA


joelsmithTT:
## Summary
Let a third-party RDMA NIC read/write Blackhole DRAM directly over PCIe.

## Approach
tt-kmd gains the ability to export a Blackhole memory aperture as a dma-buf file descriptor.  Userspace passes that fd to the NIC via the standard RDMA call (`ibv_reg_dmabuf_mr`).  The kernel's RDMA layer and the NIC driver do the rest.

## Scope
- In: Blackhole; NIC as the bus master (both read and write); dma-buf export of TLB windows; tests.
- Out: Blackhole-as-master DMA; Wormhole, anything NIC-specific.

## Milestones
1. **Verbs harness** — RDMA loopback test on Soft-RoCE (no special HW).  Validates the userspace path end-to-end with host memory.
2. **dma-buf exporter** — new tt-kmd ioctl + dma-buf support, with a self-test that runs in CI without a NIC. (No target hardware needed.)
3. **Integration** — on a Galaxy: register the exported aperture on the NIC and confirm NIC-driven read and write to Blackhole DRAM.
4. **Ship** — multi-bank coverage, tests, ABI docs, example code.

Milestones 1–2 need no special hardware and are the bulk of the work.

## Key design decision
`dma_map_resource` vs `pci_p2pdma`

## Known limitations / risks
- **Reset:** behavior is undefined if a device is reset while an app holds an exported dma-buf fd. tt-kmd will not block reset; apps must release the fd first.

alewycky-tenstorrent:
What is a memory aperture? Is it a BAR range, or is it a fixed internal address range (i.e. NOC+local)?

joelsmithTT:
BAR range (TLB window)

PR #251, opened 2026-06-08
Add TENSTORRENT_IOCTL_EXPORT_TLB_DMABUF

joelsmithTT:
Add `TENSTORRENT_IOCTL_EXPORT_TLB_DMABUF`, which exports an allocated TLB window's BAR aperture as a dma-buf fd. Handed to a peer device's driver, it lets that device do peer-to-peer PCIe DMA straight into/out of the window, which routes onto the NOC per the window's `CONFIGURE_TLB` setup.

- **ABI**: new ioctl + `struct tenstorrent_export_tlb_dmabuf`.
- **Exporter**: maps the aperture into the importer's DMA domain via `dma_map_resource()`.
- **Build**: gated on the dynamic dma-buf interface (Linux 5.8+); older kernels return `-EOPNOTSUPP` but still build.
- **Test tool**: `tools/rdma/nic_bh_p2p_dma` demonstrates and verifies a NIC <-> Blackhole DMA round trip over RoCE.

### Known limitations / follow-ups

1. ~**Importer mapping is not revocable.** If the device is reset or removed while an importer holds a mapping of the exported dma-buf, behavior is undefined.  There is no revoke today. Planned follow-up: `dma_buf_move_notify()`.~ Fixed

2. ~**`FREE_TLB` ignores outstanding exports.** `FREE_TLB` only refuses a window with an active VMA mapping; it does not track dma-buf exports, so a window can be freed (then reallocated/reconfigured) while a dma-buf still references its aperture.  Planned follow-up: a per-TLB export refcount so `FREE_TLB` returns `-EBUSY` while an export is live, released in the dma-buf `.release` callback.~ Fixed

3. **`dma_map_resource()` instead of `pci_p2pdma`.** This implementation uses the former.  It does not go through the PCI P2PDMA framework.  Planned follow-up: alternate implementation based on P2PDMA.

https://github.com/tenstorrent/tt-kmd/issues/248

### DIFF
--- a/device.h
+++ b/device.h
@@ -10,6 +10,7 @@
 #include <linux/cdev.h>
 #include <linux/reboot.h>
 #include <linux/kref.h>
+#include <linux/refcount.h>
 #include <linux/rwsem.h>
 #include <linux/wait.h>
 #include <linux/workqueue.h>
@@ -63,6 +64,7 @@ struct tenstorrent_device {
 
 	DECLARE_BITMAP(tlbs, TENSTORRENT_MAX_INBOUND_TLBS);
 	u32 tlb_counts[MAX_TLB_KINDS];	// Per-device TLB counts (may differ from dev_class defaults)
+	refcount_t tlb_refcount[TENSTORRENT_MAX_INBOUND_TLBS];
 
 	struct mutex iatu_mutex;
 	struct tenstorrent_outbound_iatu_region outbound_iatus[TENSTORRENT_MAX_OUTBOUND_IATU_REGIONS];
--- a/device.h
+++ b/device.h
@@ -77,6 +77,9 @@ struct tenstorrent_device {
 	// The stored value is arch-specific: WH stores a BAR4 sysreg offset,
 	// BH stores a raw CSM address for NOC reads.
 	u64 telemetry_tag_cache[TELEM_TAG_CACHE_SIZE];
+
+	struct list_head dmabuf_exports;
+	struct mutex dmabuf_export_lock;
 };
 
 struct tlb_descriptor;
--- a/ioctl.h
+++ b/ioctl.h
@@ -27,6 +27,7 @@
 #define TENSTORRENT_IOCTL_CONFIGURE_TLB		_IO(TENSTORRENT_IOCTL_MAGIC, 13)
 #define TENSTORRENT_IOCTL_SET_NOC_CLEANUP		_IO(TENSTORRENT_IOCTL_MAGIC, 14)
 #define TENSTORRENT_IOCTL_SET_POWER_STATE		_IO(TENSTORRENT_IOCTL_MAGIC, 15)
+#define TENSTORRENT_IOCTL_EXPORT_TLB_DMABUF		_IO(TENSTORRENT_IOCTL_MAGIC, 16)
 
 // For tenstorrent_mapping.mapping_id. These are not array indices.
 #define TENSTORRENT_MAPPING_UNUSED		0
@@ -407,6 +408,40 @@ struct tenstorrent_power_state {
 #define TT_POWER_FLAG_TENSIX_ENABLE     (1U << 2) /* 1=Enable Tensix, 0=Clock Gate Tensix */
 #define TT_POWER_FLAG_L2CPU_ENABLE      (1U << 3) /* 1=Enable L2CPU,  0=Clock Gate L2CPU */
 	__u16 power_settings[14];
+};
+
+/**
+ * TENSTORRENT_IOCTL_EXPORT_TLB_DMABUF - export a TLB window as a dma-buf
+ *
+ * Wraps a previously-allocated TLB window (see TENSTORRENT_IOCTL_ALLOCATE_TLB)
+ * in a dma-buf and returns a file descriptor for it. The fd can be handed to
+ * another device's driver -- e.g. an RDMA NIC via ibv_reg_dmabuf_mr() -- so
+ * that device can perform peer-to-peer PCIe DMA into/out of the window, which
+ * routes onto the NOC according to the window's configuration (CONFIGURE_TLB).
+ *
+ * The exported region is [offset, offset + size) within the window. A size of
+ * 0 means the remainder of the window starting at offset.
+ *
+ * WARNING: behavior is undefined if the device is reset while an importer holds
+ * a mapping of the returned dma-buf. The importer's mapping cannot be revoked.
+ * Applications must release importer mappings before resetting the device.
+ *
+ * @argsz: Must be sizeof(struct tenstorrent_export_tlb_dmabuf).
+ * @flags: Reserved for future use, must be 0.
+ * @tlb_id: A TLB window id returned by TENSTORRENT_IOCTL_ALLOCATE_TLB.
+ * @fd: OUT: the dma-buf file descriptor.
+ * @offset: Byte offset within the window at which the export begins; must be
+ *          page-aligned.
+ * @size: Number of bytes to export; 0 means to the end of the window. A
+ *        non-zero size must be a multiple of the page size.
+ */
+struct tenstorrent_export_tlb_dmabuf {
+	__u32 argsz;
+	__u32 flags;
+	__u32 tlb_id;
+	__s32 fd;
+	__u64 offset;
+	__u64 size;
 };
 
 #endif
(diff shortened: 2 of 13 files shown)


joelsmithTT:
Some example test output:

Client:
```
$ ./nic_bh_p2p_dma -r rocep201s0f0 10.32.34.129
local:  qpn=0x401 gid_index=3
path MTU = 4096 bytes
client: connected to server (qpn=0x401, bh rkey=0x400000f)
DMA write (NIC->BH)   256 MiB in   22.160 ms  ( 11.28 GiB/s)
DMA read  (BH->NIC)   256 MiB in   22.159 ms  ( 11.28 GiB/s)
client: round-trip verified (src and dst identical)
```

Server:
```
$ ./nic_bh_p2p_dma -r rocep201s0f0 -b /dev/tenstorrent/5
server: BH /dev/tenstorrent/5 window->(x=17,y=12,addr=0x0) 256 MiB exported
local:  qpn=0x401 gid_index=3
server: bh_mr rkey=0x400000f
waiting for client on tcp/18515 ...
path MTU = 4096 bytes
server: connected to client (qpn=0x401)
server: client reported DMA round-trip verified

NIC <-> Blackhole DMA round-trip verified over the fabric.
```

joelsmithTT:
Known issue: importers that pin the export rather than implementing move_notify (RDMA NICs without on-demand paging — e.g. bnxt_re, vs. mlx5 which does the dynamic/revocable path) can't be revoked, so resetting the device while such a mapping is live can generate instability if subsequent DMA is attempted.

If this turns out to bite users, we can investigate special-casing a reset block while a pinned export is live.

ioctl.h
```diff
@@ -409,4 +410,36 @@ struct tenstorrent_power_state {
...
+ * WARNING: behavior is undefined if the device is reset while an importer holds
+ * a mapping of the returned dma-buf. The importer's mapping cannot be revoked.
+ * Applications must release importer mappings before resetting the device.
+ *
+ * @argsz: Must be sizeof(struct tenstorrent_export_tlb_dmabuf).
+ * @flags: Reserved for future use, must be 0.
+ * @tlb_id: A TLB window id returned by TENSTORRENT_IOCTL_ALLOCATE_TLB.
+ * @fd: OUT: the dma-buf file descriptor.
+ * @offset: Byte offset within the window at which the export begins.
+ * @size: Number of bytes to export; 0 means to the end of the window.
```
alewycky-tenstorrent: Do these have to be page-aligned. Please document if so.

memory.c
```diff
@@ -985,6 +987,253 @@ long ioctl_configure_tlb(struct chardev_private *priv,
...
+	if (!attach->peer2peer)
+		return -EOPNOTSUPP;
+
+	return 0;
+}
+
+// A scatterlist entry's length (sg_dma_len) is a 32-bit unsigned int, so a
+// single entry cannot describe a >=4 GiB segment (4 GiB truncates to 0).
+// Split the contiguous BAR mapping across entries kept well under 4 GiB.
+#define TT_TLB_DMABUF_SG_CHUNK ((size_t)1 << 30)	// 1 GiB
```
alewycky-tenstorrent: Use SZ_1G from linux/sizes.h .

memory.c
```diff
@@ -985,6 +987,253 @@ long ioctl_configure_tlb(struct chardev_private *priv,
...
+
+static int tt_tlb_dmabuf_pin(struct dma_buf_attachment *attach)
+{
+	// The exported region is a fixed PCI BAR aperture; it never migrates.
+	return 0;
+}
+
+static void tt_tlb_dmabuf_unpin(struct dma_buf_attachment *attach)
+{
+}
```
alewycky-tenstorrent, line 1116: dma_buf_(un)map_attachment checks whether pin is supported (and needed) before using it, so I prefer that these are not defined.

device.h
```diff
@@ -60,8 +60,10 @@ struct tenstorrent_device {
 	// an in-flight FW message cannot race cleanup_hardware.
 	struct delayed_work power_down_work;
 
+	struct mutex tlb_mutex; // Serializes TLB alloc/free and per-window refcounts.
 	DECLARE_BITMAP(tlbs, TENSTORRENT_MAX_INBOUND_TLBS);
 	u32 tlb_counts[MAX_TLB_KINDS];	// Per-device TLB counts (may differ from dev_class defaults)
+	unsigned int tlb_refcount[TENSTORRENT_MAX_INBOUND_TLBS];
```
alewycky-tenstorrent: Use refcount_t.

device.h
```diff
@@ -60,8 +60,10 @@ struct tenstorrent_device {
 	// an in-flight FW message cannot race cleanup_hardware.
 	struct delayed_work power_down_work;
 
+	struct mutex tlb_mutex; // Serializes TLB alloc/free and per-window refcounts.
```
alewycky-tenstorrent: I think this mutex is unnecessary.

We have to care about these paths:
- allocate tlb ioctl
- free tlb ioctl
- export dmabuf ioctl
- chardev release
- dmabuf release

free tlb ioctl and export dmabuf ioctl ~take priv->mutex, then~ check that the device fd owns the tlb. That locks them out of any action until allocate tlb ioctl marks the device fd as owning the tlb. allocate tlb ioctl marks device fd as owning tlb as the last thing.

chardev release can't race against an ioctl, kernel refcounting prevents it.

dmabuf release can race against all of the above, but only on the tlb it owns.
- It can't really race allocate tlb ioctl because allocate will choose a different tlb.
- It can race free tlb. One of them will see the refcount go to 0 and clear the bit.
- The race with export dmabuf ioctl is uninteresting because the refcount must be at least 2. (one for allocate, one for the dmabuf) So this isn't a final tlb free anyway.
- It can race chardev release. Again one of them will see the refcount go to 0 and clear the bit.
- It can race against itself. One of them will see the refcount go to 0 and clear the bit.

test/dmabuf_export.cpp
```diff
@@ -0,0 +1,350 @@
...
+{
+    tenstorrent_export_tlb_dmabuf exp{};
+    exp.argsz = sizeof(exp);
+    exp.tlb_id = tlb_id;
+    exp.offset = offset;
+    exp.size = size;
+
+    errno = 0;
+    int ret = ioctl(fd, TENSTORRENT_IOCTL_EXPORT_TLB_DMABUF, &exp);
+    if (ret == 0 && out_fd)
```
alewycky-tenstorrent: out_fd == NULL is nonsensical and leaks the dmabuf. Don't check for it, just store to out_fd.

memory.c
joelsmithTT: In a subsequent commit, `tt_tlb_dmabuf_pin` checks whether the `tt_tlb_dmabuf` has been revoked and fails with -ENODEV in this case.

In spite of [this comment](https://elixir.bootlin.com/linux/v6.8/source/include/linux/dma-buf.h#L120), if pin is provided, unpin [must be provided too](https://elixir.bootlin.com/linux/v6.8/source/drivers/dma-buf/dma-buf.c#L626).

ctaylorTT:
@joelsmithTT thanks for adding support for dmabuf!

joelsmithTT:
Commit `Fix EXPORT_TLB_DMABUF vs RESET_DEVICE` has been reworked to outright block chip reset if there are dma-buf exports, on the grounds that:
* the RDMA NICs we have deployed in the fleet do not support on-demand paging (so the tt-kmd cannot revoke the dma-buf prior to reset)
* allowing reset to proceed during active P2P DMA is an empirically-demonstrated stability hazard requiring manual intervention (via system BMC) for recovery

memory.c
```diff
@@ -1249,13 +1278,51 @@ long ioctl_export_tlb_dmabuf(struct chardev_private *priv, struct tenstorrent_ex
...
+
+bool tenstorrent_has_tlb_dmabuf_exports(struct tenstorrent_device *tt_dev)
+{
+	bool any;
+
+	mutex_lock(&tt_dev->dmabuf_export_lock);
+	any = !list_empty(&tt_dev->dmabuf_exports);
+	mutex_unlock(&tt_dev->dmabuf_export_lock);
+
+	return any;
```
alewycky-tenstorrent: Strictly speaking this doesn't need the lock.

PR #252, opened 2026-06-10
Fix stale cdev routing for container opens after PCI re-probe

nbuncicTT:
Container devtmpfs inodes cache i_cdev on first open and never refresh it. After a PCI remove/re-probe cycle (e.g. galaxy warm reset), the cached i_cdev still points to the old detached cdev. New opens bind to the zombie device and all ioctls return -ENODEV.

When tt_cdev_open detects a detached device, it now recovers the ordinal from the minor number and redirects to the current live device via xarray lookup. Bare metal should be unaffected — udev creates fresh inodes that never hit the old cdev.

### DIFF
--- a/chardev.c
+++ b/chardev.c
@@ -730,6 +730,19 @@ static int tt_cdev_open(struct inode *inode, struct file *file)
 	struct chardev_private *private_data;
 	bool power_aware = file->f_flags & O_APPEND;
 
+	// Container devtmpfs inodes cache i_cdev from the first open and never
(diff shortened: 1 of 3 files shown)


joelsmithTT:
I'm not going to merge this since the underlying issue is solved by preventing hotplug, however this did make me look at tenstorrent_pci_remove vs tt_cdev_release more closely. I think the driver has some lingering bugs: https://github.com/tenstorrent/tt-kmd/issues/256

PR #260, opened 2026-06-28
Add NOC read/write ioctls for small reset-safe transfers 

### LINKED ISSUE #171: [Feature]: ioctl for small reset-safe transfers
### Is your feature request related to a problem? Please describe.

Non-ML device management tools want IO that is safe in the face of device reset and are willing to give up throughput and latency to get it.  

### Describe the Solution You'd Like

ioctls for 1/2/4/8 byte NOC read & write.

### Describe Alternatives You've Considered

Requiring the app to handle SIGBUS when it accesses the zapped MMIO mapping.

### Why is this Feature Important?

Long-running telemetry apps should be robust. Robustness requires simplicity.

### Proposed Design/Technical Details (Optional)

Target address is X+Y+local address. Consider also future hardware without X & Y coordinates.

Address must be size-aligned.

KMD isn't able to check the validity of the address. 

### Use Cases

UMD is the direct user. It would use this to provide simple reset-safe operations.

### Additional Context

Pavle earlier proposed NOC read/write32 ioctls.

alexapostolu:
Add NOC read and write ioctls for 1/2/4/8 byte NOC accesses. Target is
NOC(x,y) plus a local address.

Resolves #171

### DIFF
--- a/test/noc_io.cpp
+++ b/test/noc_io.cpp
@@ -1,0 +1,205 @@
+// SPDX-FileCopyrightText: © 2026 Tenstorrent Inc.
+// SPDX-License-Identifier: GPL-2.0-only
+
+// Tests for the small reset-safe NOC transfer ioctls
+// (TENSTORRENT_IOCTL_NOC_READ / TENSTORRENT_IOCTL_NOC_WRITE).
+
+#include <sys/ioctl.h>
+#include <unistd.h>
+
+#include <array>
+#include <cerrno>
+#include <cstdint>
+#include <random>
+
+#include "ioctl.h"
+
+#include "devfd.h"
+#include "enumeration.h"
+#include "test_failure.h"
+#include "tlbs.h"
+
+namespace
+{
+
+static std::mt19937_64 RNG{std::random_device{}()};
+
+uint64_t random_aligned_address(uint64_t maximum, uint64_t alignment)
+{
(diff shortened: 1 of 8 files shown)


test/noc_io.cpp
```diff
@@ -0,0 +1,205 @@
...
+    if (read_x != x || read_y != y)
+        THROW_TEST_FAILURE("NOC_READ node id mismatch");
+
+    // The same register, read at narrower widths, must agree with the low
+    // bytes of the 32-bit value.
+    if ((noc_read(fd, x, y, 0, 1, node_id_addr) & 0xff) != (node_id & 0xff))
+        THROW_TEST_FAILURE("NOC_READ 1-byte node id mismatch");
+
+    if ((noc_read(fd, x, y, 0, 2, node_id_addr) & 0xffff) != (node_id & 0xffff))
+        THROW_TEST_FAILURE("NOC_READ 2-byte node id mismatch");
```
joelsmithTT, line 82: I don't think we can expect < 32-bit accesses to work for registers.
But this ought to work for SRAM.

joelsmithTT:
I’ve picked this work up and rebased it onto current main, preserving Alex’s authorship. I also incorporated my review feedback made a few changes.  I’m going to close this draft and open a replacement PR from `joelsmith/noc-ioctls`.

PR #265, opened 2026-07-07
Treat suspend/resume as a reset

### LINKED ISSUE #253: Examine suspend support
The driver contains support for suspend, but this support is half-baked because device state loss is not propagated to users.

[This test](https://gist.github.com/joelsmithTT/238f344b3fb11af9935324d48b656967) demonstrates that TLB register state does not survive a suspend/resume cycle.  Decide on a contract:

* suspend = session lost, same as reset: bump the reset generation counter, zap mmaps, or...
* refuse suspend while in use

joelsmithTT:
Suspend loses hardware state, but the driver did nothing to tell userspace, leaving pre-suspend fds running against a device whose state had silently been wiped on resume.  Until we can save & restore that state, make suspend behave like a reset.

Closes https://github.com/tenstorrent/tt-kmd/issues/253

### DIFF
--- a/chardev.c
+++ b/chardev.c
@@ -202,7 +202,7 @@ static void bump_reset_gen(struct chardev_private *priv)
 // process may never reach, leaking windows until the pool is exhausted.
 // Caller holds reset_rwsem exclusive and has already called bump_reset_gen()
 // and tenstorrent_vma_zap().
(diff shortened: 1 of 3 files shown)


alewycky-tenstorrent:
Wormhole functional test uses rtcwake to reset the board in some cases, and BAR mmap zap will break it.

PR #266, opened 2026-07-07
Add asynchronous SMC messaging interface

### LINKED ISSUE #187: [Feature]: SMC message submission ioctl
### Is your feature request related to a problem? Please describe.

The problem is that SMC ("ARC") message submissions can conflict when applications have direct access to the queues. We have multiple queues, but it's never enough to give each possible concurrent application its own queue and we don't have a synchronization mechanism. (Even if we did, queues could be mismanaged or left in an improper state.)

### Describe the Solution You'd Like

Apps will post messages through the KMD which becomes the owner of the queue. The interface will be driven through ioctls. KMD will take the request and provide a response. User-mode polling is acceptable, at least initially.

### Describe Alternatives You've Considered

App-to-app locking, either using the KMD locks or any other OS-provided locks. Always has scope problems, how do you make sure that everybody agrees on the mapping of lock to device. What if some processes are inside a container and some are outside?

Queue state can be corrected simply by reinitializing it before each submission. (FW doesn't cache the read/write pointers.)

### Why is this Feature Important?

We need multi-process (multi-thread) safety.

### Proposed Design/Technical Details (Optional)

https://tenstorrent.atlassian.net/wiki/spaces/syseng/pages/537723135/ARC+FW+Messages

### Use Cases

All software that uses SMC messages should switch to kernel-managed messages.

### Additional Context

Performance concerns. I believe that messages are not performance-critical today, although they are used for PCIe DMA. The proposed implementation only allows one message per device fd and does nothing to pipeline messages in hardware. None of this is baked into the architecture.

Polling vs sleep. Non-performance users may prefer to sleep rather than poll for results. Sleeping is not part of the design, but it can be added on BH which supports MSI.

Shared queue for KMD and app messages. The current design uses the KMD queue for all messages. This means that the KMD has to wait for the current application message to complete before it can submit its own message. But that's already true.

joelsmithTT:
Closes https://github.com/tenstorrent/tt-kmd/issues/187
Context: https://tenstorrent.atlassian.net/wiki/spaces/syseng/pages/537723135/ARC+FW+Messages

### DIFF
--- a/chardev.c
+++ b/chardev.c
@@ -194,12 +194,39 @@ static long ioctl_get_driver_info(struct chardev_private *priv,
 	return 0;
 }
 
+// Reset reinitializes the FW queue, dropping every queued/in-flight request.
+// Return all owners to IDLE and clear the queue so nothing later acts on a
+// destroyed message: a close()-time abandon becomes a no-op, and the surviving
+// resetter's POLL sees nothing outstanding.
+static void arc_msg_reset_scrub(struct tenstorrent_device *tt_dev)
+{
+	struct chardev_private *priv, *tmp;
+
+	mutex_lock(&tt_dev->arc_msg_mutex);
+
+	list_for_each_entry_safe(priv, tmp, &tt_dev->arc_msg_queue, arc_msg.queue_node) {
+		list_del_init(&priv->arc_msg.queue_node);
+		priv->arc_msg.state = CHARDEV_MSG_IDLE;
+	}
+
+	// An abandoned in-flight owner may already have been freed by close()
+	// (see arc_msg_abandon), so don't dereference it; we drop the pointer
+	// unconditionally below either way.
+	if (tt_dev->arc_msg_inflight && !tt_dev->arc_msg_inflight_abandoned)
+		tt_dev->arc_msg_inflight->arc_msg.state = CHARDEV_MSG_IDLE;
+	tt_dev->arc_msg_inflight = NULL;
+	tt_dev->arc_msg_inflight_abandoned = false;
+
+	mutex_unlock(&tt_dev->arc_msg_mutex);
+}
+
 // Bump the device's reset generation and bring this fd's open_reset_gen along
 // with it. Other fds become permanently invalid, but the resetter keeps a live
 // fd so it can complete the reset sequence without a close/reopen window.
 static void bump_reset_gen(struct chardev_private *priv)
 {
 	priv->open_reset_gen = atomic_long_inc_return(&priv->device->reset_gen);
+	arc_msg_reset_scrub(priv->device);
 }
 
 // Reclaim TLB windows from fds this reset invalidated.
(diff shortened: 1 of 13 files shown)


msgqueue.c
```diff
@@ -8,125 +8,294 @@
...
+	struct arc_msg drain;
+	u32 queue_base;
+	u32 num_entries;
+	unsigned long timeout;
+	int ret;
+
+	if (!cls->arc_msg_locate_queue || !cls->arc_msg_trigger)
+		return -EOPNOTSUPP;
+
+	mutex_lock(&tt_dev->arc_msg_mutex);
```
alewycky-tenstorrent: Take this after the num_entries check.

msgqueue.c
```diff
@@ -8,125 +8,294 @@
...
+	if (num_entries == 0) {
+		ret = -EOPNOTSUPP;
+		goto out;
+	}
+
+	// Let any outstanding user message finish so we don't consume its
+	// response, then discard anything stale left behind by a prior exchange.
+	arc_msg_flush_inflight(tt_dev, queue_base, num_entries);
+	while (arc_msg_try_pop(tt_dev, &drain, queue_base, num_entries) == 0)
+		;
```
alewycky-tenstorrent: This loop needs a limiter of some sort, maybe num_entries.

chardev.c
```diff
@@ -621,6 +648,88 @@ static long ioctl_set_power_state(struct chardev_private *priv, struct tenstorre
...
+
+		fw_error = (priv->arc_msg.buf.header != 0);
+		memcpy(data.message, &priv->arc_msg.buf, sizeof(data.message));
+		priv->arc_msg.state = CHARDEV_MSG_IDLE;
+		mutex_unlock(&tt_dev->arc_msg_mutex);
+
+		if (copy_to_user(arg, &data, sizeof(data)) != 0)
+			return -EFAULT;
+
+		return fw_error ? -EREMOTEIO : 0;
```
alewycky-tenstorrent: Return 0 on firmware error, kmd will only return an error if the message isn't accepted by firmware.

msgqueue.c
```diff
@@ -8,125 +8,294 @@
...
+	// the FW queue is busy and we cannot submit the next message.
+	if (tt_dev->arc_msg_inflight) {
+		ret = arc_msg_try_pop(tt_dev, &resp, queue_base, num_entries);
+		if (ret == -EAGAIN)
+			return 0;
+
+		if (ret != 0) {
+			memset(&resp, 0, sizeof(resp));
+			resp.header = ARC_MSG_STATUS_HW_ERROR;
+		}
```
alewycky-tenstorrent: This should be propagated as the eventual poll ioctl returning an errno.

msgqueue.c
```diff
@@ -8,125 +8,294 @@
...
+			return 0;
+
+		list_del_init(&next->arc_msg.queue_node);
+
+		if (ret != 0) {
+			// Can't submit; complete with a failure so POLL reports it.
+			next->arc_msg.buf.header = ARC_MSG_STATUS_HW_ERROR;
+			next->arc_msg.state = CHARDEV_MSG_COMPLETED;
+			return 0;
+		}
```
alewycky-tenstorrent, line 268: This should be propagated as the eventual poll ioctl returning an errno.

alewycky-tenstorrent:
Please create a stress test that submits on multiple fds with concurrent submitters on each fd.

Do we have a FW message that e.g. echos a parameter so that you can match requests & responses?

alewycky-tenstorrent:
Reminder for us to decide whether to add slow message feedback.

alewycky-tenstorrent:
> Reminder for us to decide whether to add slow message feedback.

We decided not to.

alewycky-tenstorrent (approved):

PR #273, opened 2026-07-16
Add tlb_mutex to break mmap_lock deadlock cycle

### LINKED ISSUE #271: Lockdep: ABBA deadlock between priv->mutex and mm->mmap_lock (PIN_PAGES vs. TLB mmap)
Running with CONFIG_PROVE_LOCKING, lockdep reports a circular locking dependency:

```
python/4596 is trying to acquire lock:
 (&private_data->mutex), at: map_tlb_window+0x158/0x400 [tenstorrent]
but task is already holding lock:
 (&mm->mmap_lock), at: vm_mmap_pgoff+0xde/0x230
-> #1 (&mm->mmap_lock):        gup_fast_fallback → pin_user_pages_fast → ioctl_pin_pages
-> #0 (&private_data->mutex):  map_tlb_window → tenstorrent_mmap → tt_cdev_mmap → do_mmap
```

Path A (TENSTORRENT_IOCTL_PIN_PAGES): ioctl_pin_pages() acquires priv->mutex, then calls pin_user_pages_fast_longterm(). The GUP fast path can fall back (gup_fast_fallback) and acquire mm->mmap_lock. Order: priv->mutex → mmap_lock.

Path B (mmap() of a TLB window): The kernel holds mm->mmap_lock when calling tt_cdev_mmap() → tenstorrent_mmap() → map_tlb_window(), which acquires priv->mutex. Order: mmap_lock → priv->mutex.

Two threads sharing an mm and a device fd can deadlock: one in PIN_PAGES holding priv->mutex and blocking on mmap_lock, the other in mmap() holding mmap_lock and blocking on priv->mutex.

joelsmithTT:
Lockdep reports an ABBA deadlock between mmap_lock and priv->mutex. PIN_PAGES holds priv->mutex while GUP may acquire mmap_lock, while mapping a TLB window acquires priv->mutex with mmap_lock already held.

Add a dedicated per-fd tlb_mutex to protect TLB ownership and serialize allocation, configuration, export, mapping, free, and reset reclaim. This removes priv->mutex from the TLB mmap path and establishes the ordering mmap_lock -> tlb_mutex -> vma_lock.

Also protect the previously unlocked CONFIGURE_TLB ownership check, preventing a window from being freed and reallocated to another fd while it is being configured.

Convert the remaining bitmap accesses as well, even where broader lifecycle locking already provides exclusion, so every priv->tlbs access follows one explicit and locally auditable locking rule.

Fixes: https://github.com/tenstorrent/tt-kmd/issues/271

### DIFF
--- a/chardev.c
+++ b/chardev.c
@@ -214,10 +214,12 @@ static void tenstorrent_reset_reclaim_tlbs(struct tenstorrent_device *tt_dev)
 		if (priv->open_reset_gen == gen)
 			continue;
 
+		mutex_lock(&priv->tlb_mutex);
 		for_each_set_bit(bitpos, priv->tlbs, TENSTORRENT_MAX_INBOUND_TLBS) {
 			tenstorrent_device_free_tlb(tt_dev, bitpos);
 			clear_bit(bitpos, priv->tlbs);
 		}
+		mutex_unlock(&priv->tlb_mutex);
 	}
 	mutex_unlock(&tt_dev->chardev_mutex);
 }
@@ -839,6 +841,7 @@ static int tt_cdev_open(struct inode *inode, struct file *file)
 	INIT_LIST_HEAD(&private_data->peer_mappings);
 	INIT_LIST_HEAD(&private_data->vma_list);
 	mutex_init(&private_data->vma_lock);
+	mutex_init(&private_data->tlb_mutex);
 
 	kref_get(&tt_dev->kref);
 	private_data->device = tt_dev;
@@ -920,8 +923,13 @@ static void tt_cdev_release_resource_locks(struct chardev_private *priv)
 static void tt_cdev_release_tlbs(struct chardev_private *priv)
 {
 	unsigned int bitpos;
-	for_each_set_bit(bitpos, priv->tlbs, TENSTORRENT_MAX_INBOUND_TLBS)
+
+	mutex_lock(&priv->tlb_mutex);
+	for_each_set_bit(bitpos, priv->tlbs, TENSTORRENT_MAX_INBOUND_TLBS) {
 		tenstorrent_device_free_tlb(priv->device, bitpos);
+		clear_bit(bitpos, priv->tlbs);
+	}
+	mutex_unlock(&priv->tlb_mutex);
 }
 
 static void tt_cdev_release_power(struct chardev_private *priv)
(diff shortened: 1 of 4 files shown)


memory.c
```diff
@@ -988,7 +988,9 @@ long ioctl_allocate_tlb(struct chardev_private *priv,
 		return -EFAULT;
 	}
 
+	mutex_lock(&priv->tlb_mutex);
 	set_bit(id, priv->tlbs);
+	mutex_unlock(&priv->tlb_mutex);
```
alewycky-tenstorrent, line 993: set_bit is atomic so the mutex isn't needed  here.

chardev.c
```diff
@@ -920,8 +923,13 @@ static void tt_cdev_release_resource_locks(struct chardev_private *priv)
...
 {
 	unsigned int bitpos;
-	for_each_set_bit(bitpos, priv->tlbs, TENSTORRENT_MAX_INBOUND_TLBS)
+
+	mutex_lock(&priv->tlb_mutex);
+	for_each_set_bit(bitpos, priv->tlbs, TENSTORRENT_MAX_INBOUND_TLBS) {
 		tenstorrent_device_free_tlb(priv->device, bitpos);
+		clear_bit(bitpos, priv->tlbs);
+	}
+	mutex_unlock(&priv->tlb_mutex);
```
alewycky-tenstorrent, line 932: (tt_cdev_release already has exclusive access to priv, but keep this in case we reuse this function somewhere else.)

memory.c
```diff
@@ -1027,25 +1029,35 @@ long ioctl_free_tlb(struct chardev_private *priv, struct tenstorrent_free_tlb __
...
+	// cannot be freed (and reallocated to another fd) in between.
+	mutex_lock(&priv->tlb_mutex);
+
+	if (!test_bit(in.id, priv->tlbs)) {
+		mutex_unlock(&priv->tlb_mutex);
 		return -EPERM;
+	}
 
-	return tenstorrent_device_configure_tlb(tt_dev, in.id, &in.config);
+	ret = tenstorrent_device_configure_tlb(tt_dev, in.id, &in.config);
```
alewycky-tenstorrent: Please write this as if/else to remove the early exit.

joelsmithTT: Fixed.

memory.c
joelsmithTT: I know - I'd considered not introducing it here, but decided on consistency (every access of priv->tlbs follows the same rule that's easy to reason about).

I'm OK with removing the lock here and replacing it with a comment.

alewycky-tenstorrent (approved):

PR #274, opened 2026-07-16
Fix chardev lifetime races during PCI removal

### LINKED ISSUE #256: Examine tenstorrent_pci_remove vs tt_cdev_release for UAF bugs
Below is an AI-generated report; scrutinize it for correctness and apply the necessary fixes (if any).

# chardev open/close vs PCI remove: UAF analysis

This documents two use-after-free / list-corruption hazards between the
character-device paths (`tt_cdev_open`, `tt_cdev_release`) and the PCI teardown
path (`tenstorrent_pci_remove`).

Background that makes most of the code safe:

- `tenstorrent_device` is reference counted by `kref`. `open` takes a ref,
  `release` and `remove` each drop one. The struct therefore stays alive while
  any fd is open.
- `reset_rwsem` (held for write across `cleanup_device`/BAR unmap in remove,
  for read across the body of release/ioctl/mmap) fences device-touching work
  against the BAR unmap.

The findings below are the cases those mechanisms do **not** cover.

---

## Finding 1: remove walks `open_fds_list` without `chardev_mutex`

`tenstorrent_pci_remove` iterates the open-fd list and calls
`tenstorrent_memory_cleanup(priv)` for each fd, but holds no lock during the
walk (`enumerate.c`). Every other accessor of `open_fds_list` holds
`chardev_mutex` (open's `list_add`, release's `list_del`, `tenstorrent_vma_zap`,
`mappings_seq_show`, `pids_proc_show`).

A `close()` (or a fresh `open()`) that arrives **after** remove releases
`reset_rwsem` runs concurrently with the unlocked walk. `list_for_each_entry_safe`
only tolerates deletions by its own thread, so the concurrent `list_del` +
`kfree(priv)` corrupts the traversal and frees a node remove is about to deref.

### Interleaving (close racing remove)

| Step | Process A: `tenstorrent_pci_remove` | Process B: `tt_cdev_release` (close) |
|------|-------------------------------------|--------------------------------------|
| 1 | `mutex_lock(chardev_mutex)`; `detached = true`; `mutex_unlock` | (fd still open) |
| 2 | `cancel_delayed_work_sync(power_down_work)` | — |
| 3 | `cleanup_hardware()` / `cleanup_telemetry()` | — |
| 4 | `down_write(reset_rwsem)` | — |
| 5 | `tenstorrent_vma_zap()`; `cleanup_device()` (unmap BARs) | — |
| 6 | `up_write(reset_rwsem)` | — |
| 7 | — | `down_read(reset_rwsem)` **succeeds** (A no longer holds it) |
| 8 | `list_for_each_entry_safe(priv, tmp, open_fds_list)` begins; caches `tmp = next` | — |
| 9 | — | runs cleanup; then `mutex_lock(chardev_mutex)` |
| 10 | reads/iterates `priv` (no `chardev_mutex` held) | `list_del(&priv->open_fd)` (unlinks the node A is walking) |
| 11 | `tenstorrent_memory_cleanup(priv)` on stale/freed node | `mutex_unlock`; `up_read(reset_rwsem)` |
| 12 | continues walk using stale `tmp` → **UAF / poisoned-pointer walk** | `kfree(priv)` |

The same corruption occurs if Process B is instead a fresh `tt_cdev_open` doing
`list_add` under `chardev_mutex` at step 10.

### Why "after up_write" is required

A close that started **before** remove's `down_write` already holds
`reset_rwsem` for read, so remove's `down_write` (step 4) blocks until that
close finishes its `list_del` (under `chardev_mutex`) and `up_read`. By the time
remove reaches the walk, that `priv` is already unlinked, so remove never
touches it. Only a close/open that acquires `reset_rwsem` **after** step 6 races
the unlocked walk.

### Fix

Wrap the walk in `mutex_lock(&tt_dev->chardev_mutex)` / `mutex_unlock`, matching
`tenstorrent_vma_zap`. Lock order stays consistent
(`chardev_mutex → priv->mutex → iatu_mutex`).

---

## Finding 2: embedded `cdev` freed via a separate `kref` during `open`

`struct cdev chardev` is embedded in `tenstorrent_device`, but the struct's
memory is owned by a separate `kref`, not by the cdev's kobject. `cdev_init`
uses `ktype_cdev_default`, whose release does not free anything, so the
`cdev_get` reference the VFS holds across `open` does **not** keep `tt_dev`
memory alive.

`tt_cdev_open` dereferences `tt_dev` (via `kref_get`) only after a sleeping
`kzalloc(GFP_KERNEL)`. If a full remove completes in that window and drops the
last reference, the `kref_get` lands on freed memory.

### Interleaving (open racing remove)

| Step | Process A: `chrdev_open` → `tt_cdev_open` | Process B: `tenstorrent_pci_remove` |
|------|-------------------------------------------|-------------------------------------|
| 1 | VFS `chrdev_open`: `cdev_get(p)` succeeds (bumps cdev kobj refcount) | — |
| 2 | calls `f_op->open` = `tt_cdev_open` | — |
| 3 | `tt_dev = inode_to_tt_dev(inode)` (pointer only, no deref) | — |
| 4 | `kzalloc(GFP_KERNEL)` — **may sleep here** | — |
| 5 | — | runs full teardown (sets `detached`, unmaps BARs, etc.) |
| 6 | — | `cdev_device_del()` (`cdev_del` drops the initial kobj ref; no-op release) |
| 7 | — | `tenstorrent_device_put()` → last ref → `tt_dev_release` → `kfree(tt_dev)` |
| 8 | `kref_get(&tt_dev->kref)` on **freed** `tt_dev` → **UAF** | — |
| 9 | later: VFS `cdev_put` touches freed embedded kobj → **UAF** | — |

This needs no other fd holding a ref at step 7 (so remove's put is the last
one), and the open to be stalled at step 4. Narrow, but reachable on
hot-unplug/unbind because remove takes no lock that open holds.

### Fix

Make the cdev/`struct device` kobject the lifetime authority: give `tt_dev->dev`
a real `release` callback that performs the final teardown and use
`get_device`/`put_device` for the long-lived reference, so the struct survives
until both the last open file and `cdev_device_del` are done. At minimum,
`kref_get` must not sit behind a sleeping allocation. This is the same
underlying issue as the `.release = NULL` device with no matching `put_device`.

joelsmithTT:
Serialize remove-time fd cleanup against concurrent open and close, and use device-model references so an in-flight open cannot access a freed embedded cdev.

Fixes: https://github.com/tenstorrent/tt-kmd/issues/256

Remaining gaps:

* cleanup_hardware() is not ordered against in-flight ioctls. Remove calls it before the reset_rwsem write-side drain, so an ioctl that passed its detached check could race hardware teardown. Latent today: both callbacks return early once detached is set, which also means orderly unbind never actually shuts down present hardware. Fix: separate physical presence from logical detachment and run cleanup_hardware() after the drain.

* Opens are not rejected during removal. admit_chardev_open() never checks detached, so a racing open can succeed against a dying device and produce a zombie fd, or block forever on chardev_excl_waitqueue (no detached term in the wake condition, and remove never wakes it). Fix: check detached under chardev_mutex including after each wait, and wake the queue during removal.

### DIFF
--- a/enumerate.c
+++ b/enumerate.c
@@ -298,9 +298,6 @@ static int tenstorrent_pci_probe(struct pci_dev *dev, const struct pci_device_id
 		return err;
 	}
 
-	// The refcount created here persists until remove.
-	kref_init(&tt_dev->kref);
-
 	tt_dev->detached = false;
 	tt_dev->needs_hw_init = true;
 	tt_dev->dev_class = device_class;
@@ -404,6 +401,8 @@ fail_init_device:
 	pci_set_drvdata(dev, NULL);
 	pci_dev_put(dev);
 	xa_erase(&tenstorrent_dev_xa, ordinal);
+	// Direct kfree: this path runs before tenstorrent_register_device,
+	// so tt_dev->dev is not yet initialized and cannot be put_device'd.
 	kfree(tt_dev);
 	pci_disable_device(dev);
 	return err;
@@ -479,23 +478,23 @@ static void tenstorrent_pci_remove(struct pci_dev *dev)
 	// If this is postponed, a subsequent probe is forced to use a different ordinal.
 	xa_erase(&tenstorrent_dev_xa, tt_dev->ordinal);
 
-	tenstorrent_device_put(tt_dev);
+	// Drop the reference from device_initialize (tenstorrent_register_device).
+	// tt_dev is freed here unless open fds or dma-buf exports still hold it.
+	put_device(&tt_dev->dev);
 }
 
-static void tt_dev_release(struct kref *tt_dev_kref) {
-	struct tenstorrent_device *tt_dev = container_of(tt_dev_kref, struct tenstorrent_device, kref);
+// dev.release callback: runs when the last reference on tt_dev->dev drops.
+// Set by tenstorrent_register_device.
+void tt_dev_release(struct device *dev)
+{
+	struct tenstorrent_device *tt_dev = container_of(dev, struct tenstorrent_device, dev);
 	struct pci_dev *pdev = tt_dev->pdev;
 
 	if (tt_dev->dev_class->reboot)
 		unregister_reboot_notifier(&tt_dev->reboot_notifier);
 
-
 	pci_dev_put(pdev);
 	kfree(tt_dev);
-}
-
-void tenstorrent_device_put(struct tenstorrent_device *tt_dev) {
-	kref_put(&tt_dev->kref, tt_dev_release);
 }
 
 static int tenstorrent_suspend(struct device *dev) {
--- a/enumerate.c
+++ b/enumerate.c
@@ -412,7 +412,7 @@ fail_init_device:
 static void tenstorrent_pci_remove(struct pci_dev *dev)
 {
 	struct tenstorrent_device *tt_dev = pci_get_drvdata(dev);
-	struct chardev_private *priv, *tmp;
+	struct chardev_private *priv;
 	u16 vendor_id;
 
 	// Tear down telemetry first.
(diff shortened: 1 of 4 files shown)


enumerate.c
```diff
@@ -462,9 +461,11 @@ static void tenstorrent_pci_remove(struct pci_dev *dev)
 	// they observe detached and return -ENODEV instead of waiting forever.
 	wake_up_interruptible(&tt_dev->resource_lock_waitqueue);
 
-	list_for_each_entry_safe(priv, tmp, &tt_dev->open_fds_list, open_fd) {
+	mutex_lock(&tt_dev->chardev_mutex);
+	list_for_each_entry(priv, &tt_dev->open_fds_list, open_fd) {
 		tenstorrent_memory_cleanup(priv);
 	}
+	mutex_unlock(&tt_dev->chardev_mutex);
```
alewycky-tenstorrent, line 468: Please put this in a separate commit.

joelsmithTT: Fixed.

alewycky-tenstorrent (approved):

PR #278, opened 2026-07-24
Report firmware cable fault at probe

### LINKED ISSUE #272: Detect and log FW cable-fault low-power mode
BH firmware (v19.8.0+, commit 6a5c8b25) added a "cable fault" mode: if the DMC reports a 0W cable power limit (12V-2x6 missing/improperly seated), the SMC clock-gates all Tensix + GDDR, skips their init, and comes up in a low-power state with only the ARC/PCIe path alive. Card still enumerates on PCIe, but to a user it may look like a bricked flash rather than a HW/cable problem.

We should detect this at probe and log it clearly so users (and us in bug reports) know the FW deliberately parked the card, not that the flash failed.

Detection (both readable over PCIe since ARC stays up):

* STATUS_ERROR_STATUS0 (SCRATCH_RAM[4]), bit 1 (INIT_STAGE_CABLE_FAULT) set = FW entered cable-fault mode.
* Corroborate via cable power limit in SCRATCH_1: (val & 0xFFFF0000) == 0xCAB10000 && (val & 0xFFFF) == 0. Magic marker absent = legacy DMC, no cable-fault feature, skip the check.

On probe, read the above and emit a clear dev_warn (something like "firmware reports 12V power cable fault; card is in low-power safe mode").

joelsmithTT:
Blackhole firmware treats a 0W cable power limit from the DMC, meaning the 12V-2x6 power cable is missing or not properly seated, as an init failure and continues in a degraded low-power mode. PCIe, the ARC message queue, and telemetry come up, so the device probes and appears healthy, but firmware skips Tensix init, the GDDR MRISC load and training, and Ethernet init, leaving the RISC-V cores in reset. Nothing above the driver can use the chip. Without a diagnostic from the driver this presents as unexplained failures further up the stack.

Read STATUS_ERROR_STATUS0 (reset unit scratch RAM 4) after the A0 state message succeeds and log the cable fault bit. The successful ARC message also establishes ordering: firmware sets the message-queue-ready bit in STATUS_BOOT_STATUS0 from init_msgqueue(), which runs after the final STATUS_ERROR_STATUS0 write, so the value we read is from this boot.

Closes: https://github.com/tenstorrent/tt-kmd/issues/272

### DIFF
--- a/blackhole.c
+++ b/blackhole.c
@@ -74,6 +74,8 @@
 #define ARC_MSG_TYPE_TEST 0x90
 #define ARC_BOOT_STATUS RESET_SCRATCH(2)
 #define ARC_BOOT_STATUS_READY_FOR_MSG 0x1
+#define ARC_ERROR_STATUS0 RESET_SCRATCH(4)
+#define ARC_ERROR_STATUS0_CABLE_FAULT BIT(1)
 
 #define IATU_BASE 0x1000	// Relative to the start of BAR2
 #define IATU_OUTBOUND 0
@@ -255,6 +257,20 @@ static u32 noc_read32(struct blackhole_device *bh, u32 x, u32 y, u64 addr, int n
 	mutex_unlock(&bh->kernel_tlb_mutex);
 
 	return val;
+}
+
+static void blackhole_report_cable_fault(struct blackhole_device *bh)
+{
+	u32 error_status0 = noc_read32(bh, ARC_X, ARC_Y, ARC_ERROR_STATUS0, 0);
+
+	if (error_status0 == U32_MAX)
+		return;
+
+	if (error_status0 & ARC_ERROR_STATUS0_CABLE_FAULT)
+		dev_err(&bh->tt.pdev->dev,
+			"Firmware reported a power cable fault (error status 0x%08x); "
+			"shut down the system and check that the 12V-2x6 power cable is fully seated\n",
+			error_status0);
 }
 
 static void noc_write32(struct blackhole_device *bh, u32 x, u32 y, u64 addr, u32 data, int noc)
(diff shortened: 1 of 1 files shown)


alexapostolu (approved):

blackhole.c
```diff
@@ -257,6 +259,17 @@ static u32 noc_read32(struct blackhole_device *bh, u32 x, u32 y, u64 addr, int n
...
 
+static void blackhole_report_cable_fault(struct blackhole_device *bh)
+{
+	u32 error_status0 = noc_read32(bh, ARC_X, ARC_Y, ARC_ERROR_STATUS0, 0);
+
+	if (error_status0 == U32_MAX)
+		return;
+
+	if (error_status0 & ARC_ERROR_STATUS0_CABLE_FAULT)
+		dev_err(&bh->tt.pdev->dev, "Firmware reported a cable fault (error status 0x%08x)\n", error_status0);
```
alewycky-tenstorrent: 1. I suggest *power* cable to clarify vs. ethernet
2. It doesn't suggest any action.

alexapostolu: Should it suggest an action or should it be higher level like tools or UMD to suggest the action?

joelsmithTT: I updated the message: "Firmware reported a power cable fault (error status 0x%08x); shut down the system and check that the 12V-2x6 power cable is fully seated"

alewycky-tenstorrent (approved):

blackhole.c
alewycky-tenstorrent: (alex: no reason to defer to tools/umd since there's no mechanism/policy issue here. the only question is whether we should suggest action or assume the user will do the right thing. i prefer to suggest action because it makes it clear that this is something the user can fix, it's not a problem with the board.)

alexapostolu (approved):

PR #279, opened 2026-07-28
Reject implausible telemetry counts

### LINKED ISSUE #275: Driver trusts unvalidated counts from firmware (unbounded loops, divide-by-zero)
## Summary

Several code paths use counts and offsets read from ARC CSM without validation. Firmware-supplied values drive loop trip counts, derived NOC addresses, and divisors. Corrupt or not-yet-initialized CSM contents can wedge the device or oops the kernel.

## Instances

1. **`blackhole_populate_telemetry_cache()` (blackhole.c) — worst case.** `num_entries` is read from CSM and used directly as a loop bound, with one `noc_read32()` per iteration. A garbage count (up to ~4.3 billion) means billions of serialized MMIO reads. Each iteration takes `kernel_tlb_mutex`, starving all other NOC access. The derived `tags_addr` reads are not range-checked against CSM, and there is no `cond_resched()` in the loop.
2. **`wormhole_populate_telemetry_cache()` (wormhole.c).** Same unbounded `num_entries` loop. It range-checks each derived address, but the trip count is still fully firmware-controlled.
3. **`arc_msg_push()` / `arc_msg_pop()` (msgqueue.c).** Divide by firmware-supplied `num_entries` (`queue_info & 0xFF`) with no zero-check. Firmware reporting 0 is a kernel divide-by-zero oops.

## Impact

- On Blackhole, `RESET_DEVICE` + `POST_RESET` re-probes telemetry while holding `reset_rwsem` for write, so a bad count blocks every other ioctl/open/mmap on the device and produces hung-task warnings. The same loop can hit at probe/module load.
- A zero queue entry count from firmware oopses the kernel on any ARC message (e.g. `SET_POWER_STATE`).

## Suggested fix

Never let firmware dictate an unbounded loop or a divisor. The driver already knows the valid CSM window (`is_range_within_csm()`, `ARC_CSM_BASE`/`ARC_CSM_SIZE`).

- Telemetry parse (both chips): require `tags_addr` to be within CSM and clamp `num_entries` to what fits in CSM; warn and cap. Optionally add `cond_resched()` for defense in depth.
- Consider gating Blackhole's post-reset telemetry re-probe on a firmware-readiness check like Wormhole's `fw_ready_work`.
- `arc_msg_push()`/`arc_msg_pop()`: reject `num_entries == 0` (and ideally apply an upper bound) after reading `queue_info`.

joelsmithTT:
Closes https://github.com/tenstorrent/tt-kmd/issues/275 (partially; instance 3 is corrected by https://github.com/tenstorrent/tt-kmd/pull/266)

### DIFF
--- a/blackhole.c
+++ b/blackhole.c
@@ -491,6 +491,8 @@ static int blackhole_populate_telemetry_cache(struct tenstorrent_device *tt_dev,
 		struct telem_cache_entry key = { .tag_id = tag_id };
 		struct telem_cache_entry *entry;
 
+		cond_resched();
+
 		if (!is_range_within_csm(addr, sizeof(u32))) {
 			dev_err(&tt_dev->pdev->dev, "Telemetry tag %u has invalid address 0x%08X\n", tag_id, addr);
 			continue;
--- a/blackhole.c
+++ b/blackhole.c
@@ -458,7 +458,8 @@ static int blackhole_populate_telemetry_cache(struct tenstorrent_device *tt_dev,
 	base_addr = noc_read32(bh, ARC_X, ARC_Y, ARC_TELEMETRY_PTR, 0);
 	data_addr = noc_read32(bh, ARC_X, ARC_Y, ARC_TELEMETRY_DATA, 0);
 
-	if (!is_range_within_csm(base_addr, 1) || !is_range_within_csm(data_addr, 1)) {
+	// The telemetry header is two u32s: a version, then an entry count.
+	if (!is_range_within_csm(base_addr, 2 * sizeof(u32)) || !is_range_within_csm(data_addr, sizeof(u32))) {
 		dev_err(&tt_dev->pdev->dev, "Telemetry not available\n");
 		return -ENODEV;
 	}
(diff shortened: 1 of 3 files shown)


blackhole.c
```diff
@@ -497,14 +477,31 @@ static int blackhole_populate_telemetry_cache(struct tenstorrent_device *tt_dev,
 	tags_addr = base_addr + 8;
 	num_entries = noc_read32(bh, ARC_X, ARC_Y, base_addr + 4, 0);
 
+	// The count comes from firmware and drives one NOC read per iteration, each
+	// of which takes kernel_tlb_mutex. An implausible count means the structure
+	// is corrupt or not yet initialized, so don't trust any of it: parsing a
+	// garbage table risks caching a bogus address for a tag we care about.
```
alewycky-tenstorrent: First sentence seems not relevant?
It's mentioned in the commit description too, am I missing why it matters that it's a NOC read holding a mutex?

alewycky-tenstorrent (approved):

alewycky-tenstorrent:
Please move the `cond_resched()` calls to a separate commit.

blackhole.c
joelsmithTT: I'll impose a higher relevancy bar on agent comments going forward.

PR #288, opened 2026-08-25
CI: track the latest upstream kernel release

joelsmithTT:
Discover the latest final release tag from the cloned kernel repository so mass-build testing advances automatically when a new upstream kernel is released.


alexapostolu (approved):

alewycky-tenstorrent:
I prefer if we actually make a commit to the repo when we change the upper bound. We can have a timer-triggered workflow that checks the latest version, test builds and updates if it passes.

I had a chat with Gemini about this a while ago: https://gemini.google.com/share/d/16otKQ9BGfYJkWuStZ8ko0hrp2BAUKBoY?usp=sharing .

PR #293, opened 2026-09-10
Document the security model

joelsmithTT:
Explain that device access grants full control of a programmable, DMA-capable accelerator. Describe translating-IOMMU protection and the trust required when translation is disabled or bypassed, both of which remain supported configurations.

Require exclusive assignment between mutually untrusting users and describe the access revocation, execution and DMA shutdown, reset, and sanitization needed before reassignment. Clarify that separate-device tenant isolation also depends on the platform.

Explain the 0666 default, show a restrictive udev rule, and distinguish the documented access model from reportable driver security flaws.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,3 +64,24 @@ make dkms-remove
 ```
 make akms-remove
 ```
+
+## Security Model
+
+A Tenstorrent device is a fully programmable, DMA-capable PCIe accelerator. Opening `/dev/tenstorrent/N` grants complete control, including programming the NOC from the host or device cores. Device-node permissions control access; the driver cannot mediate subsequent device activity.
+
+**Device and host.** Without a translating IOMMU, any device user can read or overwrite arbitrary host memory. With translation, DMA is limited to memory mapped into the device's IOMMU domain. Keep the IOMMU enabled and translating to protect host memory. Disabling it (for example, with `intel_iommu=off`) or enabling passthrough (`iommu=pt`) leaves host memory unprotected from device users. The driver supports these configurations, but all device users must then be trusted with unrestricted host-memory access.
+
+**Users of one device.** All users share the device's cores, memory, and IOMMU domain. Any user can read or overwrite another's device memory and mapped host buffers, including those of privileged processes. Any user can also hang or reset the device. Mutually untrusting users must never use the same device concurrently.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -64,3 +64,24 @@ make dkms-remove
 ```
 make akms-remove
 ```
+
+## Security Model
+
+A Tenstorrent device is a fully programmable, DMA-capable PCIe accelerator. Opening `/dev/tenstorrent/N` grants complete control of it: the process can program the NOC directly, from the host or from code it loads onto the device's cores. The driver cannot mediate what the device does. The device node's permissions are the only access control.
+
+**Device and host.** Behind a translating IOMMU, the device can reach only host memory the driver has mapped for it. Without one, or with the IOMMU in passthrough mode (`iommu=pt`), it can reach all of host physical memory, and so can any user who can open it. *Suggestion: run the device behind a translating IOMMU.*
```
inmcm, line 72: I would flip it clearly state the risk up front here with the mitigation as a followup Flip it  to something like `By default, the device can reach into arbitrary host memory and perform DMA operations. Behind an  interposing MMU (such as the IOMMU), the device can reach only host memory the driver has mapped for it.`

inmcm, line 72: For the suggested mitigation, what is the practical advice?  Is it running with `iommu=foo` or other  PCIE configurations you can enable on the host?

README.md
```diff
@@ -64,3 +64,24 @@ make dkms-remove
 ```
 make akms-remove
 ```
+
+## Security Model
+
+A Tenstorrent device is a fully programmable, DMA-capable PCIe accelerator. Opening `/dev/tenstorrent/N` grants complete control of it: the process can program the NOC directly, from the host or from code it loads onto the device's cores. The driver cannot mediate what the device does. The device node's permissions are the only access control.
```
inmcm, line 70: Is there any additional concern if you have more than one PCIe device in a host? Does the host driver all communications between them by default or could you run a multi-tenant system with N+1 cards out of the box?

README.md
joelsmithTT: The practical advice is basically "don't disable it or put it in passthrough mode".  I updated the language to be more clear about what the driver's position on this is.

README.md
joelsmithTT: Separate devices can be assigned to different users through device-node permissions, but those permissions control who can open each device; they do not establish isolation between the devices.

The driver does not mediate device-originated communication.  Isolation depends on the platform’s IOMMU configuration, PCIe peer-to-peer routing/isolation, and any direct links between devices.  Merely having separate cards does not establish separate security domains.

For mutually untrusting tenants, that needs a validated deployment configuration.  Internally we use PCIe passthrough to VMs for isolation, but I would bring in our cloud team for a supported deployment recommendation rather than promise that N cards provide N isolated tenants.

inmcm (approved):

PR #294, opened 2026-09-19
Groundwork for moving reset polling into the kernel

joelsmithTT:
The longer-term goal is for POST_RESET to wait for both the reset marker and firmware readiness, removing the need for userspace to sleep or poll between USER_RESET and POST_RESET.  That change is not included here.

Work toward it meant comparing the initialization paths used by probe and POST_RESET. That exposed several assumptions hidden by the userspace delays, as well as duplicated telemetry setup.  Following the newly exposed failure paths through probe also found places where the driver continued with unreadable hardware or saved invalid reset state.

* Probe Wormhole telemetry synchronously.  The delayed worker was a workaround for slow GDDR training in older firmware; current firmware no longer needs it.  Probe now waits for a bounded period instead.
* Wait for firmware before init_hardware communicates with it.  Previously, probe or POST_RESET could send messages before ARC firmware was listening and still report success.  Both architectures now perform a bounded wait at the start of init_hardware.
* Treat all-ones reads as an init_hardware failure.  This leaves a hung device available only for identification and reset.  POST_RESET also keeps needs_hw_init set until reinitialization succeeds.
* Register PCIe performance counters through the device rather than telemetry.  They do not depend on firmware and are now managed through dev.groups.
* Use the same telemetry setup path from probe and POST_RESET. This removes duplicated per-architecture setup and cleanup.  It also means that telemetry missing at probe can be registered after a successful reset.
* Fail probe when PCI config space is unreadable rather than continuing with an unresponsive device.
* Save MPS only after init_hardware confirms that the NOC is reachable. Reset recovery now skips restoring MPS if no valid value was saved.


slindsayTT:
> Probe Wormhole telemetry synchronously. The delayed worker was a workaround for slow GDDR training in older firmware; current firmware no longer needs it. Probe now waits for a bounded period instead.

How old is said firmware? Do we need to be concerned about backward compatibility?

joelsmithTT:
> > Probe Wormhole telemetry synchronously. The delayed worker was a workaround for slow GDDR training in older firmware; current firmware no longer needs it. Probe now waits for a bounded period instead.
> 
> How old is said firmware? Do we need to be concerned about backward compatibility?

With this KMD patch + WH Galaxy with CMFW 19.6.0 or older, the driver will not publish telemetry.  This should not block the ability to update the firmware.

slindsayTT:
At the moment, WH f/w is stuck at 19.6.0 because of a DDR change that caused a problem in WH Galaxy.
See https://tenstorrent.atlassian.net/browse/SYS-5072

